Issue #193, opened 2020-08-06
Skaffold integration


marcinc:
The scope of this integration is to auto create a skeleton of a skaffold manifest and automate its configuration, so that the local (skaffold) development flows are enabled for the project at the time Kev project is initiated.

PR #226, opened 2020-08-27
Prepare kev project for Skaffold

marcinc:
Addresses: #193 

Note, this first pass doesn't pre-configure skaffold manifest with `Build` phase artifacts. As an extension to this, we could ask the user to specify locations of Dockerfiles for respective compose project components, with an option to skip some/all if upstream images are used.

Note:

`kev init --skaffold` only generates skaffold.yaml if it doesn't already exist. We don't want to potentially break the existing valid manifest. Kev manifest now contains additional `skaffold` field telling us whether the project uses Skaffold or not. Generated Skaffold manifest contains a profile for each defined environment + some additional convenience profiles. 

`kev render` updates skaffold.yaml profiles with the last rendered manifests paths for each rendered environment. Currently it's the only mutation we perform.


pkg/kev/kev.go
```diff
@@ -28,24 +29,35 @@ import (
...
-
+	ManifestName       = "kev.yaml"
 	defaultEnv         = "dev"
 	configFileTemplate = "docker-compose.kev.%s.yaml"
 )
 
 // Init initialises a kev manifest including source compose files and environments.
 // A default environment will be allocated if no environments were provided.
-func Init(composeSources, envs []string, workingDir string) (*Manifest, error) {
+func Init(composeSources, envs []string, workingDir string, skaffold bool) (*Manifest, *SkaffoldManifest, error) {
```
ezodude, line 40: See comments regarding a `kev.PrepareForSkaffold(...)` above ^.

pkg/kev/skaffold.go
```diff
@@ -48,6 +50,57 @@ func NewSkaffoldManifest(envs []string) (*SkaffoldManifest, error) {
...
+func LoadSkaffoldManifest(path string) (*SkaffoldManifest, error) {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	var s *SkaffoldManifest
+	return s, yaml.Unmarshal(data, &s)
+}
+
+// UpdateSkaffoldProfiles uptadets skaffold profiles with appropriate kubernetes files output paths
```
ezodude, line 63: `uptadets` -> `updates`

pkg/kev/skaffold_test.go
```diff
@@ -136,4 +136,57 @@ var _ = Describe("Skaffold", func() {
...
+				outputPath := "testdata" // point at any existing directory for test!
+
+				envToOutputPath := map[string]string{
+					envName: outputPath,
+				}
+
+				It("updates the matching profile with new manifests path selecting all the files in that directory", func() {
+					manifest.UpdateProfiles(envToOutputPath)
+					Expect(manifest.Profiles[0].Deploy.KubectlDeploy.Manifests).To(ContainElement(filepath.Join(outputPath, "*")))
+				})
```
ezodude, line 163: It would be great if `manifest` has a convenience method to return the value in `manifest.Profiles[0].Deploy.KubectlDeploy.Manifests`.

ezodude (commented):
Skaffold FTW! 🤘

Some minor design and refactoring suggestions.

cmd/kev/cmd/init.go
```diff
@@ -88,25 +92,53 @@ func runInitCmd(cmd *cobra.Command, _ []string) error {
 		return displayInitError(err)
 	}
 
-	manifest, err := kev.Init(files, envs, ".")
+	manifest, skaffoldManifest, err := kev.Init(files, envs, ".", skaffold)
 	if err != nil {
 		return displayInitError(err)
 	}
```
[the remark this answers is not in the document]
marcinc: Funnily enough it started like this but since we map command to a matching action I encapsulated everything in Init. I don't necessarily like it and yeah I think I revert it back to what it was originally...

pkg/kev/skaffold_test.go
marcinc: there is currently no need for it but might add some as and when required.

ezodude (approved):
🚀

Issue #229, opened 2020-09-07
Change default storage class to empty string


marcinc:
When experimenting locally default storage class might not be set in Kubernetes cluster. Storage class shouldn't be set on a volume unless explicitly specified.

PR #231, opened 2020-09-07
PV storage class fix

marcinc:
Addresses #229


ezodude (commented):
Looks good.

Should we update the `/testdata/...` fixtures to not include the `standard` storage class? Most fixtures assume it exists by default.

marcinc:
> Should we update the /testdata/... fixtures to not include the standard storage class? Most fixtures assume it exists by default.

No, we don't need to change testdata assets. Storage class is a valid parameter so if someone specifies one because they use it then that's OK.

ezodude (approved):
👍

PR #232, opened 2020-09-08
Fix liveness probe command interpolation

### LINKED ISSUE #230: Fix default liveness probe to never include CMD part
Currently Kev sets default liveness probe command if not detected in source compose file, however, the CMD part set by the docker-compose is also taken into account which breaks the deployment.

marcinc:
Resolves #230


ezodude (approved):
🚀

PR #233, opened 2020-09-08
Reconcile: overridden environment overlay settings are lost.

ezodude:
This resolves https://github.com/appvia/kev-planning/issues/29.

- Refactored algorithm as previous implementation was becoming awkward to patch.
- Fixed up to either rollback or keep environment overlay settings.
- Added new tests to cover new scenarios.
- Removed dependency on the `r3labs/diff` package.


cmd/compose/main.go
```diff
@@ -0,0 +1,97 @@
...
+
+func main() {
+	files := []string{
+		"docker-compose.yaml",
+		"docker-compose.kev.dev.yaml",
+	}
+	fmt.Println("Parsing ./docker-compose.yaml")
+
+	p, err := loadFromSources(files)
+	// p, err := rawProjectFromSources(files)
```
marcinc: this line not needed I guess?

marcinc (approved):

PR #237, opened 2020-09-08
Environment override files names match those of the first source comp…

### LINKED ISSUE #222: Environment overrides name to match source compose files
Environment compose files naming should match that of first passed source compose file to make it easier to find them.

Example:

```sh
kev init -f compose.yaml -e prod
```

should produce the following prod env override file:

```sh
compose.kev.prod.yaml
```

Currently, prefix is fixed at `docker-compose.kev.`.

marcinc:
…ose file

Resolves #222


ezodude (approved):
👍

PR #239, opened 2020-09-09
Removes explicit env_file handling

### LINKED ISSUE #204: Refactor render logic and remove unnecessary bits e.g. env_file related
Render logic was largely refactored to work with composego Project struct. This means that some elements are handled internally by the compose-go library. One example of such element is env_file - it gets expanded automatically and key-value pairs from environment file are automatically injected to the project service `Environment`. 

This story is to clean up / remove redundant code.

marcinc:
This PR removes explicit env_file handling as compose-go library automatically loads all environment variables defined in env_file and appends any explicitly defined env variables. compose-go discards information about the env_file used and so there is currently no need for any special env_file handling. All compose project service env variables are automatically expanded in the Kubernetes deployment definition.

Resolves #204

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/compose-spec/compose-go v0.0.0-20200907084823-057e1edc5b6f
 	github.com/google/go-cmp v0.5.2
 	github.com/imdario/mergo v0.3.11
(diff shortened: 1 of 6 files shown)


ezodude (approved):
💯 Deleting code - lovely 😺

PR #254, opened 2020-09-11
Toggle project services

### LINKED ISSUE #207: Add ability to toggle project services on/off per environment
At the moment all the environment specific configuration overlays are used to control the generated manifests, however, there is no way to control what goes in that specific environment in terms of components. 

Add ability to control component presence per env.  

This could be potentially just another label which then informs business logic to exclude / include the component when generating manifests.

marcinc:
Resolves #207

### DIFF
--- a/pkg/kev/converter/kubernetes/project_service.go
+++ b/pkg/kev/converter/kubernetes/project_service.go
@@ -34,6 +34,17 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+// enabled returns Bool telling Kev whether app component is enabled/disabled
+func (p *ProjectService) enabled() bool {
+	if val, ok := p.Labels[config.LabelComponentEnabled]; ok {
+		if val == "false" {
+			return false
+		}
+	}
+
+	return true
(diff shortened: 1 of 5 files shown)


pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -34,6 +34,17 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+// enabled returns Bool telling Kev whether app component is enabled/disabled
+func (p *ProjectService) enabled() bool {
+	if val, ok := p.Labels[config.LabelComponentEnabled]; ok {
+		if val == "false" {
+			return false
+		}
```
ezodude, line 42: This can introduce some subtle issues, e.g. misspelling false will trigger that the service is `enabled`.

IMO, we should parse the value to ensure we have a bool and act accordingly. Maybe something like `strconv.ParseBool()` can help?

ezodude (commented):
Make sense. Left an implementation comment.

ezodude (approved):

PR #258, opened 2020-09-11
Add support for Pod fields in environment variables

### LINKED ISSUE #182: Add support for Pod fields in environment variables

marcinc:
Resolves #182

### DIFF
--- a/pkg/kev/converter/kubernetes/transform.go
+++ b/pkg/kev/converter/kubernetes/transform.go
@@ -1151,37 +1151,65 @@ func (k *Kubernetes) configEnvs(projectService ProjectService) ([]v1.EnvVar, err
 			v = &temp
 		}
 
-		// @step check whether env var value references secret or configmap
-		// e.g. `secret.my-secret-name.my-key`, `config.my-config-name.config-key`
+		// @step generate EnvVar spec and handle special value reference cases for `secret`, `configmap` or `pod` field
+		// e.g. `secret.my-secret-name.my-key`, `config.my-config-name.config-key`, `pod.metadata.namespace`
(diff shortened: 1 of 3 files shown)


pkg/kev/converter/kubernetes/transform.go
```diff
@@ -1151,37 +1151,65 @@ func (k *Kubernetes) configEnvs(projectService ProjectService) ([]v1.EnvVar, err
...
+			paths := []string{
+				"metadata.name", "metadata.namespace", "metadata.labels", "metadata.annotations",
+				"spec.nodeName", "spec.serviceAccountName", "status.hostIP", "status.podIP", "status.podIPs",
+			}
+
+			path := strings.Join(parts[1:], ".")
+
+			sort.Strings(paths)
+			i := sort.SearchStrings(paths, path)
+			if i < len(paths) && paths[i] == path {
```
ezodude, line 1196: I get how you're using `sort.SearchStrings` to find a `path` in `paths`.

Perhaps you can use a simple `Set` to to do the same without having to resort to `sort` and `SearchStrings` ... for e.g. 
```
paths := map[string]bool{
  "metadata.name": true,
 ...
}
...
if paths[path] {
...
}
```

ezodude (commented):
Minor tweak

pkg/kev/converter/kubernetes/transform.go
marcinc: going to turn original string detection in slice into a helper function as it is used in a couple of places now. sort.SearchStrings seem much more elegant than building a map first to find something in it later.

marcinc: ^ Done in the follow up PR: #259

ezodude (approved):
👍

PR #260, opened 2020-09-14
Add profiles to existing skaffold manifest when initialising with --skaffold

### LINKED ISSUE #250: Add/update skaffold profiles for externally created skaffold manifest file.
When skaffold was created externally i.e. not bootstrapped by Kev, we need to be able to still work with that and inject kev profiles into it. This should allow us to automate some of the dev flows at later stage.  

Need to be mindful of profile names as they may clash with what user already defined so might need to rename profile names generated by Kev to ensure there is no conflicts.

marcinc:
Resolves #250

### DIFF
--- a/cmd/kev/cmd/init.go
+++ b/cmd/kev/cmd/init.go
@@ -122,40 +122,8 @@ func runInitCmd(cmd *cobra.Command, _ []string) error {
 		// set skaffold path in kev manifest
 		manifest.Skaffold = kev.SkaffoldFileName
 
-		if manifestExistsForPath(skaffoldManifestPath) {
-			// skaffold manifest already present - add additional profiles to it!
-			// Note: kev will skip profiles with names matching those of existing
-			// profile names defined in skaffold to avoid profile "hijack".
-
-			updatedSkaffold, err := kev.AddProfiles(skaffoldManifestPath, envs, true)
-			if err != nil {
-				return displayInitError(err)
-			}
-			if err := writeTo(skaffoldManifestPath, updatedSkaffold); err != nil {
-				return displayInitError(err)
-			}
-
-			results = append(results, skippableFile{
-				FileName: kev.SkaffoldFileName,
-				Updated:  true,
-			})
-
-		} else {
-
-			skaffoldManifest, err := kev.PrepareForSkaffold(envs)
-			if err != nil {
-				return displayInitError(err)
-			}
-
-			if skaffoldManifest != nil {
-				if err := writeTo(kev.SkaffoldFileName, skaffoldManifest); err != nil {
-					return displayInitError(err)
-				}
-
-				results = append(results, skippableFile{
-					FileName: kev.SkaffoldFileName,
-				})
-			}
+		if err := createOrUpdateSkaffoldManifest(skaffoldManifestPath, envs, &results); err != nil {
+			return displayInitError(err)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/kev/cmd/init.go
```diff
@@ -118,25 +119,43 @@ func runInitCmd(cmd *cobra.Command, _ []string) error {
...
 			})
+
+		} else {
+
+			skaffoldManifest, err := kev.PrepareForSkaffold(envs)
+			if err != nil {
+				return displayInitError(err)
+			}
+
+			if skaffoldManifest != nil {
```
ezodude, line 150: Not sure why this check required, as the lines before already do the work.

cmd/kev/cmd/init.go
```diff
@@ -118,25 +119,43 @@ func runInitCmd(cmd *cobra.Command, _ []string) error {
...
 
-		skaffoldManifest, err := kev.PrepareForSkaffold(manifest, skaffoldManifestPath, envs)
-		if err != nil {
-			return displayInitError(err)
-		}
+		// set skaffold path in kev manifest
+		manifest.Skaffold = kev.SkaffoldFileName
 
-		// don't override existing skaffold.yaml
 		if manifestExistsForPath(skaffoldManifestPath) {
```
ezodude, line 125: So this is a `createOrUpdateSkaffoldManifest` type setup. Refactoring it into a `func` will help make it more obvious.

pkg/kev/kev.go
```diff
@@ -49,15 +48,12 @@ func Init(composeSources, envs []string, workingDir string) (*Manifest, error) {
 }
 
 // PrepareForSkaffold initialises a skaffold manifest for kev project.
-// It'll also set the Skaffold field in kev manifest with skaffold file path passed as argument.
-func PrepareForSkaffold(manifest *Manifest, skaffoldPath string, envs []string) (*SkaffoldManifest, error) {
+func PrepareForSkaffold(envs []string) (*SkaffoldManifest, error) {
```
ezodude, line 51: ❤️ this simplification.

pkg/kev/skaffold.go
```diff
@@ -60,6 +71,22 @@ func LoadSkaffoldManifest(path string) (*SkaffoldManifest, error) {
 	return s, yaml.Unmarshal(data, &s)
 }
 
+// AddProfiles injects kev profiles to existing Skaffold manifest
+// Note, if profile name already exists in the skaffold manifest then profile won't be added
+func AddProfiles(path string, envs []string, includeAdditional bool) (*SkaffoldManifest, error) {
```
ezodude, line 76: Looking around the code `includeAdditional` is always `true` ... and `NewSkaffoldManifest`  runs `AdditionalProfiles()`.

Is there a case where we wouldn't want to run `AdditionalProfiles()`?

ezodude (commented):
All in all makes sense.

pkg/kev/skaffold.go
marcinc: This is useful primarily in testing. Might add additional flag to control whether to include additional (helper) profiles - which currently are included by default - to the manifest in the future PR.

cmd/kev/cmd/init.go
marcinc: old artefact. removed.

ezodude (approved):
👍

PR #262, opened 2020-09-15
Regression: Reconcile should not append a new service env vars onto an overlay. 

### LINKED ISSUE #261: Regression: Reconcile should not append a new service env vars onto an overlay.
Env vars should only be overridden in overlays.

This regression adds them by default when a new service is added.

ezodude:
Small fix, resolves #261


marcinc (approved):
LG

PR #263, opened 2020-09-16
Condense kev env override config to reduce repetition.

### LINKED ISSUE #251: Condense kev env override config to reduce repetition.
Proposal based on Spike:

**On Init**

- Parse source compose files.
- Extract all Labels.
- Condense to a set of Labels for every found `Service` and `Volume`.
- Save`docker-compose.kev.*.yaml` to disk, it will contain the condensed set if Labels only.

**On Reconcile**

- Parse source compose files.
- Load Overlay per environment.
- Condense the set of labels in the Parsed Sources to match the Labels in the Overlay.
- Now reconcile Overlay to Sources.
- Save`docker-compose.kev.*.yaml` to disk, it will contain the result of the reconciliation with condensed Labels for NEW services or Volumes.

**On Render**

- Parse source compose files.
- Load Overlay per environment.
- Expand Overlay Labels, with Defaults for missing Labels.
- Expand Overlay Labels, if label `kev.service.type` is missing, then update with value of from sources.
- Merge Env Into Sources (as before).

ezodude:
Resolves #251 

Progress:

- [x] Init: condense env overlay labels.
- [x] Reconcile: reconcile based on condensed env overlay labels.
- [x] Render: render based on condensed env overlay labels.
- [x] Reconcile: new added services and volumes should have condensed env overlay labels.


pkg/kev/envs.go
```diff
@@ -69,12 +69,13 @@ func (e *Environment) GetServices() Services {
...
 // GetService retrieves the specific service by name from the environment's overlay.
 func (e *Environment) GetService(name string) (ServiceConfig, error) {
-	for _, s := range e.GetServices() {
-		if s.Name == name {
-			return s, nil
-		}
-	}
-	return ServiceConfig{}, fmt.Errorf("no such service: %s", name)
+	return e.overlay.getService(name)
+	// for _, s := range e.GetServices() {
```
marcinc, line 73: is this code needed?

pkg/kev/envs.go
```diff
@@ -111,12 +112,13 @@ func (e *Environment) VolumeNames() []string {
...
 // GetVolume retrieves a specific volume by name from the environment's overlay volumes.
 func (e *Environment) GetVolume(name string) (VolumeConfig, error) {
-	for k, v := range e.GetVolumes() {
-		if k == name {
-			return v, nil
-		}
-	}
-	return VolumeConfig{}, fmt.Errorf("no such volume: %s", name)
+	return e.overlay.getVolume(name)
+	// for k, v := range e.GetVolumes() {
```
marcinc, line 116: is it needed?

pkg/kev/envs.go
```diff
@@ -69,12 +69,13 @@ func (e *Environment) GetServices() Services {
 
 // GetService retrieves the specific service by name from the environment's overlay.
 func (e *Environment) GetService(name string) (ServiceConfig, error) {
-	for _, s := range e.GetServices() {
-		if s.Name == name {
-			return s, nil
-		}
-	}
-	return ServiceConfig{}, fmt.Errorf("no such service: %s", name)
+	return e.overlay.getService(name)
```
marcinc, line 72: a small suggestion. perhaps we should use `override` instead of `overlay` to be more explicit as to what it is? `overlay` is a bit ambiguous. Obviously same comment apply to all instances below.

pkg/kev/envs.go
ezodude: Thanks, its dead code.

pkg/kev/envs.go
ezodude: Thanks, its dead code.

pkg/kev/envs.go
ezodude: Actually this matches docker-composes [own docs](https://docs.docker.com/compose/extends/).

marcinc (dismissed):
LG

marcinc (approved):

PR #264, opened 2020-09-17
Validate overlay config settings to ensure they match expected input values.

### LINKED ISSUE #257: Validate overlay config settings to ensure they match expected input values.
When loading environment override overlays from disk, ensure the overlays have valid settings.

- [x] Investigate [gojsonschema](https://github.com/xeipuuv/gojsonschema) to see if it helps (used by compose-go).

ezodude:
Resolves #257 

Still work in progress.


pkg/kev/config/schema.go
```diff
@@ -0,0 +1,60 @@
...
+		LabelWorkloadCPU:                       map[string]interface{}{"type": "string"},
+		LabelWorkloadImagePullPolicy:           map[string]interface{}{"type": "string", "enum": []string{"IfNotPresent", "Always"}},
+		LabelWorkloadLivenessProbeCommand:      map[string]interface{}{"type": "string"},
+		LabelWorkloadLivenessProbeDisabled:     map[string]interface{}{"type": "string", "enum": []string{"false", "true"}},
+		LabelWorkloadLivenessProbeInitialDelay: map[string]interface{}{"type": "string", "format": "duration"},
+		LabelWorkloadLivenessProbeInterval:     map[string]interface{}{"type": "string", "format": "duration"},
+		LabelWorkloadLivenessProbeRetries:      map[string]interface{}{"type": "string", "pattern": `^\d+$`},
+		LabelWorkloadLivenessProbeTimeout:      map[string]interface{}{"type": "string", "format": "duration"},
+		LabelWorkloadMaxCPU:                    map[string]interface{}{"type": "string"},
+		LabelWorkloadMaxMemory:                 map[string]interface{}{"type": "string", "pattern": `^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$`},
```
marcinc, line 35: I guess we shouldn't allow P :) T is also questionable. Maybe we should allow up to G? Is this regex going to cover cases like `Gi` or `Mi`? Same comment applies to the line below.

pkg/kev/config/schema.go
```diff
@@ -0,0 +1,60 @@
...
+	},
+	"additionalProperties": false,
+}
+
+var VolumesSchema = map[string]interface{}{
+	"$schema": "http://json-schema.org/draft/2019-09/schema#",
+	"$id":     "http://appvia.io/schemas/kev-service-labels-schema.json",
+	"type":    "object",
+	"properties": map[string]interface{}{
+		LabelVolumeSize:         map[string]interface{}{"type": "string", "pattern": `^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$`},
```
marcinc, line 54: Comment as above.

pkg/kev/config/schema.go
```diff
@@ -0,0 +1,60 @@
...
+
+package config
+
+var ServicesSchema = map[string]interface{}{
+	"$schema": "http://json-schema.org/draft/2019-09/schema#",
+	"$id":     "http://appvia.io/schemas/kev-service-labels-schema.json",
+	"type":    "object",
+	"properties": map[string]interface{}{
+		LabelComponentEnabled:                  map[string]interface{}{"type": "string", "enum": []string{"false", "true"}},
+		LabelServiceType:                       map[string]interface{}{"type": "string", "enum": []string{"None", "Headless", "ClusterIP", "Nodeport", "LoadBalancer"}},
```
marcinc, line 25: Is validator going to be case insensitive? We should support mixed cases e.g. `headless` == `HeAdLESs` etc..

marcinc (commented):
Like the added validation bits. Left a few comments.

pkg/kev/config/schema.go
marcinc: Although it makes sense to allow T or even P in this case.

pkg/kev/config/schema.go
ezodude: Nope, for now, it is case sensitive. But it will flag the expected values on error, e.g. `error ... expected:  "None", "Headless", "ClusterIP", "Nodeport", etc..`.

On a first look, supporting mixed case with JSON schema seems to be a bit of a hassle,
- **Proposal from 2013**: https://groups.google.com/g/json-schema/c/V-ZAJfV7mt8
- **Ugly**: https://stackoverflow.com/questions/60001238/how-to-make-the-json-schema-flexible-enough-to-handle-case-sensitive-input-for-e

Maybe leave that for another ticket and see if users find it annoying?

pkg/kev/config/schema.go
ezodude: This regex is quite permissive - lifted from K8s.
- Happy to remove P & T.
- And, yes it handles `Mi` and `Gi`.

pkg/kev/config/schema.go
ezodude: Same as above.

pkg/kev/config/schema.go
marcinc: Technically it's OK, so could leave as is and if someone requests a few Petabytes of memory it's unlikely to fit on any host anyway...

pkg/kev/config/schema.go
marcinc: Sure, can leave as is for now. Maybe check the docs if it matches.

pkg/kev/config/schema.go
ezodude: Left the regex as is.

pkg/kev/config/schema.go
ezodude: @marcin Rolled back update and left the regex as is.

marcinc (approved):
LG

PR #266, opened 2020-09-18
Add build artifacts section in skaffold manifest when initializing Kev with --skaffold

### LINKED ISSUE #265: Skaffold - detect and add artefact definition (build step)
Figure out a way to bootrap build config based on the application structure. This is something skaffold init attempts to do i.e. detect application structure by scanning for Dockerfiles and linking them with their respective application components.

As Kev now can also bootstrap skaffold configuration we could replicate, or invoke, what skaffold init would normally do.

Need to mindful of the fact that skaffold also prompts user where decision need to be made about Dockerfile association to a specific component.

Also, there is no point in building external services docker images as they are just consumed by the application stack.

marcinc:
Resolves #265


pkg/kev/skaffold_test.go
```diff
@@ -262,4 +274,49 @@ var _ = Describe("Skaffold", func() {
...
+			// example: src/myservice/Dockerfile will result in `myservice` service name
+
+			Context("and detected remote registry image names matching service name", func() {
+				analysis := &kev.Analysis{
+					Dockerfiles: []string{"src/myservice/Dockerfile"},
+					Images:      []string{"quay.io/myorg/myservice", "myservice"},
+				}
+
+				It("picks remote registry image path and sets correct Build configuration", func() {
+					skaffoldManifest.SetBuildArtifacts(analysis)
```
ezodude, line 298: More of a style thing, but can `skaffoldManifest.SetBuildArtifacts(analysis)` be wrapped in a `JustBeforeEach` block? 

This might mean you have to declare `analysis` next to `var skaffoldManifest ...` and use it in the different specs.

pkg/kev/skaffold.go
```diff
@@ -262,6 +284,88 @@ func (s *SkaffoldManifest) AdditionalProfiles() {
...
+func (s *SkaffoldManifest) SetBuildArtifacts(analysis *Analysis) error {
+
+	// don't set build artefacts if no analysis available
+	if analysis == nil {
+		return nil
+	}
+
+	buildArtifacts := map[string]string{}
+
+	for _, d := range analysis.Dockerfiles {
```
ezodude, line 297: Would be cool if this block is factored out into its own func `e.g. collectImagesArtefacts(...) or something similar`. 

Helps communicate what's happening without having to trawl through the code.

pkg/kev/skaffold.go
```diff
@@ -262,6 +284,88 @@ func (s *SkaffoldManifest) AdditionalProfiles() {
...
+			if found := re.FindAllStringSubmatchIndex(image, -1); found != nil {
+				buildArtifacts[context] = image
+				break
+			} else {
+				buildArtifacts[context] = svcNameFromContext
+			}
+		}
+	}
+
+	artifacts := []*latest.Artifact{}
```
ezodude, line 318: Would be cool if this block (L:318 - L:327) is factored out into a func `e.g. setupImageArtefacts(...)`.

ezodude (commented):
Makes sense ... minor reviews.

pkg/kev/skaffold_test.go
marcinc: yeh, multiple ways of going about this and often adding JustBeforeEach add extra code achieving the same thing. Also it's just two cases am testing here so no biggie. If it results in less / same number of statements I'll do it otherwise leave as is.

pkg/kev/skaffold_test.go
```diff
@@ -277,39 +277,49 @@ var _ = Describe("Skaffold", func() {
...
+				BeforeEach(func() {
+					analysis = &kev.Analysis{
+						Dockerfiles: []string{"src/myservice/Dockerfile"},
+						Images:      []string{"quay.io/myorg/myservice", "myservice"},
+					}
+				})
 
 				It("picks remote registry image path and sets correct Build configuration", func() {
-					skaffoldManifest.SetBuildArtifacts(analysis)
+					// skaffoldManifest.SetBuildArtifacts(analysis)
```
ezodude, line 307: ^ Dead code?

ezodude (commented):
👍  ... spotted commented code.

ezodude (approved):
👍

Issue #270, opened 2020-09-23
Add "Kev Dev" mode watching source compose, reconciling & re-rendering K8s manifests


marcinc:
Add `dev` mode for Kev. This command would watch source compose files for any changes and continuously reconcile all the changes and re-render Kubernetes manifests for specified environments. If none envs are provided it'll automatically render default `dev` environment.

The purpose for this command is to automate iteration process and allow for source changes to be automatically reflected in corresponding k8s manifests. 

This should also work for cases where Skaffold is used.

- [x] watch source compose files
- [x] reconcile source changes with environment overrides
- [x] re-render k8s manifests for specified environments, or `dev` if none supplied.
- [x] update skaffold manifest to point at the latest k8s manifests location for each environment profile

PR #271, opened 2020-09-23
Add "dev" command for continuous compose files changes reconciliation and re-rendering of K8s manifests

marcinc:
Addresses #270


pkg/kev/kev.go
```diff
@@ -123,3 +124,72 @@ func Render(format string, singleFile bool, dir string, envs []string) error {
 
 	return nil
 }
+
+// Dev continuously watches source compose files and rebuilds Kubernetes manifests on each change.
+// Note: It'll rebuild manifessts for all tracked environments, unless list of environments specifically provided.
+func Dev(envs []string, change chan<- string) error {
+	workDir, err := os.Getwd()
```
ezodude, line 131: It would be great if we can pass in the workDir as a param.

It makes testing easier and helps us use the func in a lib context.

pkg/kev/kev.go
```diff
@@ -123,3 +124,72 @@ func Render(format string, singleFile bool, dir string, envs []string) error {
 
 	return nil
 }
+
+// Dev continuously watches source compose files and rebuilds Kubernetes manifests on each change.
+// Note: It'll rebuild manifessts for all tracked environments, unless list of environments specifically provided.
```
ezodude, line 129: Based on the implementation so far, `kev.Dev` just notifies the caller of changes to the manifest's sources.

It's up to the client to react to those changes - which I think makes a ton of sense.

**How about we rename `kev.Dev` to `kev.Watch`** and not pass in the `envs` param?

That way any complications from dealing with `envs` are up to the caller to deal with.

marcinc: changed

pkg/kev/kev.go
marcinc: Watch func just takes a list of files to watch and a channel. Caller decides what needs to be watched.

pkg/kev/kev.go
ezodude: @marcinc after reviewing the latest changes, I think there was a bit of misunderstanding - prob because I mentioned `any complications from dealing with envs` ... my bad.

We should catch up quickly after stand up.

pkg/kev/kev.go
```diff
@@ -125,8 +125,20 @@ func Render(format string, singleFile bool, dir string, envs []string) error {
 	return nil
 }
 
-// Watch continuously watches specified files and notifies changes to a channel
-func Watch(files []string, change chan<- string) error {
+// Watch continuously watches source compose files & environment overrides and notifies changes to a channel
+func Watch(envs []string, change chan<- string) error {
+	workDir, err := os.Getwd()
```
ezodude, line 130: All makes sense, except it made sense to make `workDir` a param as before.

ezodude (approved):
👍

PR #273, opened 2020-09-23
Detect potential secrets in envs or env_files

### LINKED ISSUE #136: Detect potential secrets in envs or env_files
As we'll need to handle local env_files at `init` and unpack them to kev configuration we might need to scan for common terms like PASSWORD, etc. and warn user when detected. 

User might refer to local env file for a set of environment variables and this could potentially contain secret info locally. Try and detect by scanning for commonly used terms and communicate to the user.

ezodude:
Resolves #136 

- [x] Found regexps to help match secrets from other OSS projects.
- [x] Retrofitted regexps as Secret Matchers to identify potential env var secret leaks.
- [x] Can detect secrets in manifest sources env vars.
- [x] Can detect secrets in environment override env vars.
- [x] Wired in the detect command to run after `init`.
- [x] Wired in the detect command to run after `Reconcile` and before `Render`.


marcinc (commented):
Need a rebase

ezodude:
@marcinc done (fyi used github directly).

pkg/kev/services.go
```diff
@@ -66,6 +69,53 @@ func (s Services) Set() map[string]bool {
 	return out
 }
 
+func (s Services) detectSecrets(matchers []map[string]string, reporter io.Writer) error {
+	var matches []secretHit
+	for _, svc := range s {
+		matches = append(matches, svc.detectSecretsInEnvVars(matchers)...)
+	}
+
+	if nothingDetected := len(matches) < 1; nothingDetected {
```
marcinc, line 78: why not just `if len(matches) == 0 { ... }` ?

pkg/kev/services.go
```diff
@@ -66,6 +69,53 @@ func (s Services) Set() map[string]bool {
...
+				" → WARNING, leak detected    ... Service [%s], env var [%s], %s\n",
+				m.svcName,
+				m.envVar,
+				m.description)))
+	}
+	return nil
+}
+
+func (sc ServiceConfig) detectSecretsInEnvVars(matchers []map[string]string) []secretHit {
+	var out []secretHit
```
marcinc, line 95: maybe call it `matches` like in the detectSecrets method above?

marcinc (approved):
2 tiny comments. LG otherwise.

pkg/kev/services.go
marcinc: or perhaps `hits`?

marcinc (approved):

PR #276, opened 2020-09-24
Add a verbose mode to Kev cmds, to communicate extra runtime info.

### LINKED ISSUE #274: Add a verbose mode to Kev cmds, to communicate extra runtime info.

ezodude:
Resolves #274 

- [x] Add `verbose` flag to the root command - similar to how `docker-compose` works.
- [x] Only display reconcile info when `verbose` flag is set.
- [x] Always warn on specific detected secret regardless of `verbose` flag.
- [x] Only display details of all secret detection (detected/non) when `verbose` flag is set.
- [x] Only display render logs when `verbose` flag is set.
- [x] Ensure how errors are displayed works with new log approach.
- [x] Fix tests and remove `reporter`.

To run in `verbose` mode:
- **init**: `kev -v init` or `kev --verbose init`.
- **render**: `kev -v render` or `kev --verbose render`.

### DIFF
--- a/cmd/kev/cmd/render.go
+++ b/cmd/kev/cmd/render.go
@@ -17,9 +17,7 @@
 package cmd
 
 import (
-	"fmt"
-	"io/ioutil"
-	"strings"
+	"os"
 
 	"github.com/appvia/kev/pkg/kev"
 	"github.com/appvia/kev/pkg/kev/log"
@@ -49,6 +47,7 @@ var renderCmd = &cobra.Command{
 			if err := fn(cmd, args); err != nil {
 				return err
 			}
+			os.Stdout.Write([]byte("\n"))
 		}
 		return nil
 	},
@@ -91,6 +90,8 @@ func init() {
 }
 
 func runRenderCmd(cmd *cobra.Command, _ []string) error {
+	cmdName := "Render"
+
 	format, err := cmd.Flags().GetString("format")
 	singleFile, err := cmd.Flags().GetBool("single")
 	dir, err := cmd.Flags().GetString("dir")
@@ -101,16 +102,12 @@ func runRenderCmd(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	reporter := getReporter(true)
-	if !verbose {
-		log.SetOutput(ioutil.Discard)
-	}
-
-	_, _ = reporter.Write([]byte("✓ Render\n"))
+	setReporting(verbose)
+	displayCmdStarted(cmdName)
+	log.DebugTitlef("Output format: %s", format)
 	if err := kev.Render(format, singleFile, dir, envs); err != nil {
 		return err
 	}
-	_, _ = reporter.Write([]byte(fmt.Sprintf(" → Rendering manifests in %s output format ... Done\n", strings.Title(format))))
 
 	return nil
 }
--- a/cmd/kev/cmd/render.go
+++ b/cmd/kev/cmd/render.go
@@ -17,6 +17,10 @@
 package cmd
 
 import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
 	"github.com/appvia/kev/pkg/kev"
 	"github.com/appvia/kev/pkg/kev/log"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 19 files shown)


cmd/kev/cmd/render.go
```diff
@@ -91,18 +95,22 @@ func runRenderCmd(cmd *cobra.Command, _ []string) error {
...
+		log.SetOutput(ioutil.Discard)
+	}
 
+	_, _ = reporter.Write([]byte("✓ Render\n"))
 	if err := kev.Render(format, singleFile, dir, envs); err != nil {
 		return err
 	}
-
-	log.Info("🧰 App render complete!")
+	_, _ = reporter.Write([]byte(fmt.Sprintf(" → Rendering manifests in %s output format ... Done\n", strings.Title(format))))
```
marcinc, line 113: When in "regular" mode this obviously missing some context e.g. which environment was rendered.

marcinc (commented):
Is there a way to make verbosity global flag position independent? Supporting something like `kev dev --verbose` would be more intuitive I think.

cmd/kev/cmd/render.go
ezodude: Good point - need re-think this.

marcinc (commented):
Shouldn't (potential) secret detection be verbose always? Same goes for other commands..

ezodude:
@marcinc Since we're close to `docker-compose`, I though it would be good to keep the same expectation.

`kev dev --verbose` style does not work on compose.

ezodude:
Pending ...

- [ ] Ensure how errors are displayed works with new log approach.
- [ ] Fix tests and remove `reporter`.

cmd/kev/cmd/render.go
```diff
@@ -45,6 +47,7 @@ var renderCmd = &cobra.Command{
 			if err := fn(cmd, args); err != nil {
 				return err
 			}
+			os.Stdout.Write([]byte("\n"))
```
marcinc, line 50: I think I've seen it a few times. Could this go in one of functions?

pkg/kev/changeset.go
```diff
@@ -174,72 +173,72 @@ func (cset changeset) HasNoPatches() bool {
...
-func (chg change) patchVersion(override *composeOverride, reporter io.Writer) {
+func (chg change) patchVersion(override *composeOverride) {
 	if chg.Type != UPDATE {
 		return
 	}
 	pre := override.Version
 	newValue := chg.Value.(string)
 	override.Version = newValue
-	_, _ = reporter.Write([]byte(fmt.Sprintf(" → version updated, from:[%s] to:[%s]\n", pre, newValue)))
+	log.DebugDetailf("version updated, from:[%s] to:[%s]", pre, newValue)
```
marcinc, line 203: Why not just Debug? is DebugDetail different from a regular debug?

ezodude: It was to differentiate between the old use of Debug ... DebugDetail uses a different prefix (` → `) from Debug.

Saying that, I don't think we use the old Debug anymore. If that's the case, then I'll change `DebugDetail` -> `Debug`.

cmd/kev/cmd/render.go
ezodude: Yup, its used twice. For now, I can't think of an elegant way to abstract this.

Its used as a cmd separator, in a context where the cmds are unaware of how they're ordered and therefore cannot assert if a new line is required or not.

TBH, the UX still needs more love. So will definitely revisit this in the future. The `logrus` lib is a bit lacking.

pkg/kev/changeset.go
ezodude: done

pkg/kev/log/log.go
```diff
@@ -258,12 +235,10 @@ func decorate(level string, f ...Fields) logrus.Fields {
 		switch level {
 		case "debug-title":
 			fields["prefix"] = successTitlePrefix
-		case "debug-detail":
+		case "debug":
 			fields["prefix"] = detailPrefix
```
marcinc, line 239: As `DebugPrefix` is now removed perhaps it makes sense to address/change other prefixes too, i.e. drop the emojis and replace with something else to be consistent?

ezodude: of course.

ezodude: IMHO, that should be work in a separate PR.

E.g. The way error logging works in the `converter` (and some places in `render`) doesn't entirely conform to the new UX.

We have a UX ticket on the backlog which that can address the above while revamping logging for our text user interface.

ezodude: @marcinc ^

marcinc (approved):

PR #280, opened 2020-09-30
Display reconcile rules, when communicating reconcile results.

ezodude:
Resolves, #275 - depends on PR https://github.com/appvia/kev/pull/276

In verbose mode, displays the rules (dimmed using the `\033[2m` text attribute):

```zsh
% kev -v render
᛬ Reconcile ...
HOW DOES RECONCILE WORK?
᛫ New services & volumes in a project will be added to all environments.
᛫ Removed services & volumes from a project will be removed from all environments.
᛫ Generally, updates to a project's settings DON'T affect environments.
᛫ Updates to a service's ports in a project DO affect environments.
᛫ ENV VARS will be removed if they are found in an environment file but not in the project.
debug ✓  Reconciling environment [dev]
debug  →  nothing to update
```

### DIFF
--- a/cmd/kev/cmd/reconcile.go
+++ b/cmd/kev/cmd/reconcile.go
@@ -58,8 +58,7 @@ func displayReconcileRules(verbose bool) {
 		_, _ = fmt.Fprintf(os.Stdout, "\033[2mHOW DOES RECONCILE WORK?\n")
 		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ New services & volumes in a project will be added to all environments.\n")
 		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Removed services & volumes from a project will be removed from all environments.\n")
-		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Generally, updates to a project's settings DON'T affect environments.\n")
-		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Updates to a service's ports in a project DO affect environments.\n")
-		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ ENV VARS will be removed if they are found in an environment file but not in the project.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Environment settings trump project settings, with the exception of ports.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ To avoid Env Var hell, an environment can only override a service's Env Vars.\n")
 	}
 }
--- a/cmd/kev/cmd/reconcile.go
+++ b/cmd/kev/cmd/reconcile.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 	"path"
 
@@ -35,6 +36,7 @@ func runReconcileCmd(cmd *cobra.Command, _ []string) error {
 
 	setReporting(verbose)
 	displayCmdStarted(cmdName)
+	displayReconcileRules(verbose)
 
 	manifest, err := kev.Reconcile(workingDir)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


cmd/kev/cmd/reconcile.go
```diff
@@ -17,34 +17,49 @@
...
 }
+
+func displayReconcileRules(verbose bool) {
+	if verbose {
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2mHOW DOES RECONCILE WORK?\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ New services & volumes in a project will be added to all environments.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Removed services & volumes from a project will be removed from all environments.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Generally, updates to a project's settings DON'T affect environments.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Updates to a service's ports in a project DO affect environments.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ ENV VARS will be removed if they are found in an environment file but not in the project.\n")
```
marcinc, line 63: Now that I think about that I'm not sure this rule makes much sense.. Could imagine that one might add env variable in the environment specific override in special cases. It would probably complicate reconciliation so could leave as is.

cmd/kev/cmd/reconcile.go
```diff
@@ -17,34 +17,49 @@
...
 
 	return nil
 }
+
+func displayReconcileRules(verbose bool) {
+	if verbose {
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2mHOW DOES RECONCILE WORK?\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ New services & volumes in a project will be added to all environments.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Removed services & volumes from a project will be removed from all environments.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Generally, updates to a project's settings DON'T affect environments.\n")
```
marcinc, line 61: Reword suggestion (combining 2 lines together) : `Updates to a project's services settings don't affect environment configurations, with exception to ports.` or  `Environment config settings takes precedence over source compose settings, with exception to ports`? Would be good to explain the impact of port change somehow?

cmd/kev/cmd/reconcile.go
ezodude: Our assumption here, is that we only allow you to override ENV VARS - not create new ones. So, as not to end in some environment variable hell. 😳

For your scenario: 
- Add the ENV VAR to your source.
- Override the target environments with the correct value.
- Override other environments with a blank value.

I should reword the rule slightly to convey what I just explained. 🙂

cmd/kev/cmd/reconcile.go
ezodude: I'll see what I can do. Need to make sure our points are short and sweet.

ezodude:
@marcinc rebased latest from master.

cmd/kev/cmd/reconcile.go
```diff
@@ -50,3 +52,13 @@ func runReconcileCmd(cmd *cobra.Command, _ []string) error {
...
 	return nil
 }
+
+func displayReconcileRules(verbose bool) {
+	if verbose {
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2mHOW DOES RECONCILE WORK?\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ New services & volumes in a project will be added to all environments.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Removed services & volumes from a project will be removed from all environments.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ Environment settings trump project settings, with the exception of ports.\n")
+		_, _ = fmt.Fprintf(os.Stdout, "\033[2m᛫ To avoid Env Var hell, an environment can only override a service's Env Vars.\n")
```
marcinc, line 62: Would suggest dropping `To avoid Env Var hell, `

marcinc (approved):

PR #286, opened 2020-10-02
Readiness probe support

### LINKED ISSUE #99: Add readiness probe support
Currently no equivalent in compose spec but could add that as a step when generating workload specs.

marcinc:
Resolves #99 

In this PR:
- [x] support for readiness probe configuration
- [x] docs improvements
- [x] minor bug fixes and tests

### DIFF
--- a/pkg/kev/converter/kubernetes/project_service.go
+++ b/pkg/kev/converter/kubernetes/project_service.go
@@ -560,7 +560,7 @@ func (p *ProjectService) healthcheck() (*v1.Probe, error) {
 		initialDelaySeconds := p.livenessProbeInitialDelay()
 		failureThreshold := p.livenessProbeRetries()
 
-		if len(command) == 0 || timoutSeconds == 0 || periodSeconds == 0 ||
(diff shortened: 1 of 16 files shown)


pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -560,7 +560,7 @@ func (p *ProjectService) healthcheck() (*v1.Probe, error) {
 		initialDelaySeconds := p.livenessProbeInitialDelay()
 		failureThreshold := p.livenessProbeRetries()
 
-		if len(command) == 0 || timoutSeconds == 0 || periodSeconds == 0 ||
+		if len(command) == 0 || len(command[0]) == 0 || timoutSeconds == 0 || periodSeconds == 0 ||
```
ezodude, line 563: Just curious: why is `len(command[0]) == 0 ` required?

ezodude (approved):
🚀

pkg/kev/converter/kubernetes/project_service.go
marcinc: Command is mapped to a string slice, so we need to ensure that at least first element is not empty, so in cases you pass in an empty string as your command validation should kick in preventing you from setting invalid probe.

PR #287, opened 2020-10-02
Make selected config options case insensitive

### LINKED ISSUE #283: Make environment specific label values case insensitive
Recently added validation enforces that values of config parameters are as defined in the reference. In practice this makes iterating on the application harder, especially when in `kev dev` mode.  

Consider allowing lowercase value equivalents in the validator or make it entirely case insensitive.

marcinc:
Resolves #283


ezodude (approved):
💯

Issue #291, opened 2020-10-02
Refine Skaffold profiles generated by Kev


marcinc:
Refine Kev injected Skaffold profiles to make it easier for developers to use. 

Currently KEV includes environment specific + helper profiles but this could be refined further to make the use of them more obvious.

PR #292, opened 2020-10-05
Refine skaffold profiles

marcinc:
Addresses #291 

- [x] Skaffold "build" artefacts should also take into account images & build context defined in Docker Compose files in case Skaffold analysis didn't detect any Images (usually due to missing k8s manifests in the project directory)
- [x] Refine Kev environment specific Skaffold profiles to not include kubectl context 
- [x] Revisit helper profiles naming convention


ezodude (approved):
👍

Issue #293, opened 2020-10-08
`kev dev` to wrap around `skaffold dev`


marcinc:
Currently `kev dev` solely focuses on watching source Compose files and automatically re-renders K8s manifests for chosen environments.

Kev can initialise the project with `--skaffold` flag which create a brand new Skaffold manifest (or update existing one) and injects custom profiles to the Skaffold config. 

Currnetly, to take advantage of Skaffold local dev pipeline, one need to start `skaffold dev` in a separate session. This creates a bit of cognitive overhead and context switching. 

It would be helpful to activate `skaffold dev` pipeline at the time of iterating on kev project (with `kev dev`). The end result would lead to the following flow:

1). `kev dev --skaffold` starts both `skaffold dev` and `kev dev` pipelines. 
2) each change to source/environment specific Compose file(s) will trigger re-render of K8s manifests.
3) Skaffold automatically detects the change in manifests and runs its build/push/deploy pipeline.

`kev dev` without `--skaffold` flag passed should behave as is i.e. just watch source Compose files and regenerate application manifests.

PR #294, opened 2020-10-08
[Experimental] Wrap around skaffold dev

marcinc:
Addresses #293 

**NOTE: This feature is EXPERIMENTAL**

This PR:

- [x] Adds option to activate `skaffold dev` loop when iterating on the project with `kev dev` with `--skaffold` flag.
- [x] Ensure that `skaffold dev` is properly configured before starting the loop:
  - [x] the right skaffold profile(s) are used
  - [x] the correct target k8s namespace is passed
  - [x] the correct kubecontext is referenced pointing skaffold at the correct cluster

Usage:

   To start iterating on the application with Kev and automatically activate skaffold dev loop, pass `--skaffold` flag to `kev dev` command. 

Skaffold will expect the following options to work:
* `--kubecontext | -k` - name of the kubecontext to be used by skaffold. This will point it to the correct K8s cluster. Defaults to current kubectl context.
* `--namespace | -n` - name of target K8s namespace to which automatic deployment should happen. Defaults to `default` if not specified.
* `--kev-env` - name of one of Kev managed environments. This will tell Skaffold which manifests to deploy. Defaults to first of specified environments (via `--environments | -e`) or default `dev` environment if none have be provided.

Breaking the dev loop with Ctrl+C will automatically cleanup deployment target namespace and remove all objects. It'll also attempt to prune locally built docker images. Be mindful, that prune operation may take a while and could fail so user must keep an eye on number and size of images built locally. Auto pruning before dev loop will be added in the future Skaffold versions and should help keep that in check.


cmd/kev/cmd/dev.go
```diff
@@ -167,9 +167,21 @@ func runDevCmd(cmd *cobra.Command, args []string) error {
 		return err
```
ezodude, line 168: To keep err handling consistent with other cmds, just use `return displayError(err)`.

Should prob apply to all `return err` in this file.

cmd/kev/cmd/dev.go
```diff
@@ -167,9 +167,21 @@ func runDevCmd(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+
 	if skaffold {
-		profileName := kevenv + kev.EnvProfileNameSuffix
-		go kev.RunSkaffoldDev(cmd.Context(), cmd.OutOrStdout(), []string{profileName}, namespace, kubecontext, 1000)
+		if skaffoldConfigPath, skaffoldConfig, ok := kev.ActivateSkaffoldDevLoop(workDir); ok {
+			if err := writeTo(skaffoldConfigPath, skaffoldConfig); err != nil {
+				log.Warnf("Couldn't write Skaffold config: %s.", err)
+				return err
```
ezodude, line 180: How about if we simply handle this as a log.Error (and use `return displayError(err)`)?

pkg/kev/kev.go
```diff
@@ -183,3 +184,41 @@ func Watch(workDir string, envs []string, change chan<- string) error {
...
+		log.Warn(`Can't activate Skaffold dev loop. Kev wasn't initialized with --skaffold.
+	If you don't currently have skaffold.yaml in your project you may bootstrap a new one with "skaffold init" command.
+	Once you have skaffold.yaml in your project, make sure that Kev references it by adding "skaffold: skaffold.yaml" in kev.yaml!`)
+		return "", nil, false
+	}
+
+	configPath := path.Join(workDir, manifest.Skaffold)
+
+	if !fileExists(configPath) {
+		log.Error(`Can't file Skaffold config file referenced by Kev manifest. Have you initialized Kev with --skaffold?
```
ezodude, line 210: `log.Error(`Can't file Skaffold` -> `log.Error(`Can't find Skaffold`

ezodude (commented):
Make sense.

Generally, feels like we might need to start thinking about a top level Plugins package. As `Skaffold` logic is starting to gently bleed into core `Kev` logic.

pkg/kev/skaffold.go
```diff
@@ -452,3 +463,116 @@ func (s *SkaffoldManifest) sortProfiles() {
...
+	runCtx, cfg, err := runContext(opts)
+
+	r, err := runner.NewForConfig(runCtx)
+
+	if err != nil {
+		log.Error("Skaffold dev failed", err)
+		return err
+	}
+
+	prune := func() {}
```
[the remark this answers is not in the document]
marcinc: Yes it's a bit confusing at first. These bits have been extracted from skaffold codebase, mind. From what I gathered it was to ensure the ordering of Pruning and Cleanup i.e. to ensure Pruning happens after Cleanup. I've now updated the code to actually make use of those when Ctrl+C is intercepted.

cmd/kev/cmd/dev.go
marcinc: updated

cmd/kev/cmd/dev.go
marcinc: updated

ezodude (approved):
👍

PR #295, opened 2020-10-08
Automate testing and validation with multiple real life docker-compose projects

ezodude:
Resolves https://github.com/appvia/kev-planning/issues/35

This PR only considers the `wordpress-mysql` docker-compose project. Including all the checks and CI integration.

**I propose to run this for a couple of days to ensure that its solid.**

Once we're happy, we can add more docker-compose projects as separate tickets. As we'll need to get familiar enough with every project to be able to craft the necessary tests.

To run the e2e tests,
```
make e2e
```

This PR,
- [x] Uses Kind for e2e tests.
- [x] Includes setup and teardown across all e2e tests.
- [x] `wordpress-mysql` e2e: verifies K8s artefacts are as expected.
- [x] `wordpress-mysql` e2e: verifies K8s artefacts are as running correctly.
- [x] `wordpress-mysql` e2e: curl localhost to check wordpress is running.
- [x] Scheduled to run every hour on CircleCI.


e2e/bin/e2e.sh
```diff
@@ -0,0 +1,93 @@
...
+export PATH=${PATH}:${PWD}/bin
+
+log()      { (2>/dev/null printf "$@${NC}\n"); }
+announce() { log "${GREEN}[$(date +"%T")] [INFO] $@"; }
+failed()   { log "${YELLOW}[$(date +"%T")] [FAIL] $@"; }
+
+usage() {
+  cat <<EOF
+  Usage: $(basename $0)
+  --build-cli    <bool>    : indicates should should build the kore cli (defaults: ${BUILD_CLI})
```
marcinc, line 41: need edit: should should -> should... kore -> kev!

e2e/helper.bash
```diff
@@ -0,0 +1,79 @@
...
+  generate-manifests
+  apply-manifests
+}
+
+# wait-on-deployment is responsible for waiting for a deployment to deploy
+wait-on-deployment() {
+  local namespace=$1
+  local labels=$2
+
+  for ((i=0; i<=60; i++)); do
```
marcinc, line 54: this could be re-worked slightly to make it timeout based - eg. give it 60 sec to complete the test, otherwise fail. If it fails due to timeout being not long enough then we could bump it. Currently we might wait up to 3 minutes for this to eventually fail.

e2e/helper.bash
```diff
@@ -77,3 +93,18 @@ ensure-service-type(){
   $KUBECTL -n ${namespace} get service -l ${label} --no-headers | grep -i ${type}
 }
 
+check-app-is-running(){
+  local namespace=$1
+  local label=$2
+  local port=$3
+
+  for ((i=0; i<=60; i++)); do
+    if eval "$KUBECTL -n ${namespace} exec $($KUBECTL get pod -n "${namespace}" -l "${label}" -o name) --  curl -sLI localhost:${port}" ; then
```
marcinc, line 102: extract pod name to separate command perhaps, to make it easier to follow.

.circleci/config.yml
```diff
@@ -11,11 +11,33 @@ jobs:
...
+  nightly:
+    triggers:
+      - schedule:
+          cron: "0 * * * *"
+          filters:
+            branches:
+              only:
+                - master
+    jobs:
+      - e2e
```
marcinc, line 43: minor: NL missing

hack/e2e/setup-kind.sh
```diff
@@ -0,0 +1,45 @@
...
+  KIND_BINARY_URL="https://github.com/kubernetes-sigs/kind/releases/download/${VERSION}/kind-linux-amd64"
+  wget -O "${KIND}" "${KIND_BINARY_URL}"
+  chmod +x "${KIND}"
+}
+
+main() {
+  # get kind
+  install_kind_release
+}
+main
```
marcinc, line 45: minor: NL missing

marcinc (commented):
left a few minor comments

e2e/bin/e2e.sh
ezodude: Ha!

e2e/helper.bash
ezodude: @marcinc as per your suggestion, I started on a `timeout` based solution.

But `timeout` is not cross platform - no timeout on Mac OSX … so, to keep things simple and portable I'd rather stick with this solution.

I'll try `sleep 2` to bring it down to 2 mins max.

marcinc: ok, leave as is for now. if we envisage "loop and wait" to be used in multiple places then maybe turn it into a separate function?

marcinc (approved):
LG

PR #298, opened 2020-10-13
`dev` command shouldn't break on invalid config

### LINKED ISSUE #290: `dev` command shouldn't break on invalid config
Currently, when in `dev` mode the watch loop stops and the program exits when a config error is detected. 

Instead, we should 
- Keep the watch loop running.
- Intercept the error.
- Notify the user.

This will enable users to make any necessary changes / fixes to the config label(s).

ezodude:
Resolves #290 

- [x] Only log config errors - but don't stop.
- [x] Formatted dev mode starting banner.

### DIFF
--- a/cmd/kev/cmd/ux.go
+++ b/cmd/kev/cmd/ux.go
@@ -36,6 +36,14 @@ func displayCmdStarted(cmdName string) {
 	_, _ = os.Stdout.Write([]byte("> " + cmdName + "...\n"))
 }
 
+func displayDevModeStarted(envs []string) {
+	_, _ = fmt.Fprintf(os.Stdout, "\033[2m[development mode] ... listing environments:\n")
+	for _, env := range envs {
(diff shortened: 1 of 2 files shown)


cmd/kev/cmd/ux.go
```diff
@@ -36,6 +36,14 @@ func displayCmdStarted(cmdName string) {
 	_, _ = os.Stdout.Write([]byte("> " + cmdName + "...\n"))
 }
 
+func displayDevModeStarted(envs []string) {
+	_, _ = fmt.Fprintf(os.Stdout, "\033[2m[development mode] ... listing environments:\n")
```
marcinc, line 40: `watched environments`?

marcinc (approved):
1 minor comment. LGTMO

cmd/kev/cmd/ux.go
ezodude: sure ... I borrowed the `skaffold` lingo ... `watched` make sense.

marcinc (approved):

PR #310, opened 2020-10-16
Infer kev.service.type as ClusterIP

### LINKED ISSUE #308: Infer kev.service.type as ClusterIP
As a developer, I want my services to default to a private discovery level, so I can have granular control on how they are discovered across environments.

Also, addresses #288

## Benefits
- Removes confusion on how and why a service type was selected by Kev.
- In local K8s development, this ensures port-forwarding is explicit as some services did not auto port forward.
- Ensures that a developer can easily follow discovery/ingress guidelines as set by a provider or parent org.

## Solution

Any docker-compose service with a port attribute will default to `kev.service.type`: `ClusterIP`.

ezodude:
Resolves #308 

- [x] Ensures any service with a port configured is mapped to `kev.service.type`:`ClusterIP`
- [x] Updates e2e tests to match new reality.
- [x] Removes redundant code.
- [x] Docs and reference updated.

### DIFF
--- a/docs/reference/config-params.md
+++ b/docs/reference/config-params.md
@@ -634,13 +634,13 @@ The `service` group contains configuration detail around Kubernetes services and
 
 ## kev.service.type
 
-Defines the type of Kubernetes service for a specific workload. See official K8s [documentation](https://kubernetes.io/docs/concepts/services-networking/service/).
+Defines the type of Kubernetes service for a specific workload. See official K8s [documentation](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types).
 
 Although Kev provides a variety of types you can use, it only tries to extract two types of services from the compose configuration, namely `None` or `ClusterIP`.
 
-If you need a different type, please configure it manually. The different types are listed and explained below. 
+If you need a different type, please configure it manually. The different types are listed and explained below. Related official K8s
 
-Here is the heuristic used to extract a `None` or `ClusterIP` type:
+Here is the heuristic used to extract a service type:
 
 * If compose project service publishes a port (i.e. defines a port mapping between host and container ports):
     * It will assume a `ClusterIP` service type
@@ -678,7 +678,7 @@ It is ideal for an internal service or locally testing an app before exposing it
 
 This service type is the most basic way to get external traffic directly to your service.
 
-Its opens a specific port on all the Nodes (the VMs), and any traffic that is sent to this port is forwarded to the service.
+Its opens a specific port on each of the K8s cluster Nodes, and any traffic that is sent to this port is forwarded to the ClusterIP service which is automatically created.
 
 You'll be able to contact the NodePort Service, from outside the cluster, by requesting `<NodeIP>:<NodePort>`.
 
@@ -690,7 +690,7 @@ This is the same as a `ClusterIP` service type, but lacks load balancing or prox
 
 Specifically, it does have a service IP, but instead of load-balancing it will return the IPs of the associated Pods.
 
-It is ideal for scenarios like Pod to Pod communication.
+It is ideal for scenarios such as Pod to Pod communication or clustered applications node discovery.
 
 #### LoadBalancer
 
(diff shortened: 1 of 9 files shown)


ezodude:
@marcinc quick question, does the check below still make sense? 

https://github.com/appvia/kev/blob/master/pkg/kev/converter/kubernetes/project_service.go#L160

marcinc:
> quick question, does the check below still make sense?

Yes, it still makes sense to leave it there.

marcinc (changes requested):
It'd be good to update reference accordingly as part of this change.

docs/reference/config-params.md
```diff
@@ -656,9 +658,47 @@ services:
...
+
+It is ideal for running a service with limited availability, e.g. a demo app.
+
+#### Headless
+
+This is the same as a `ClusterIP` service type, but lacks load balancing or proxying. Allowing you to connect to a Pod directly.
+
+Specifically, it does have a service IP, but instead of load-balancing it will return the IPs of the associated Pods.
+
+It is ideal for scenarios like Pod to Pod communication.
```
marcinc, line 693: ... `such as Pod to Pod communication or clustered applications node discovery.` ?

docs/reference/config-params.md
```diff
@@ -656,9 +658,47 @@ services:
...
+
+In development, you can access this service on your localhost using [Port Forwarding](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/).
+
+It is ideal for an internal service or locally testing an app before exposing it externally.
+
+#### Nodeport
+
+This service type is the most basic way to get external traffic directly to your service.
+
+Its opens a specific port on all the Nodes (the VMs), and any traffic that is sent to this port is forwarded to the service.
```
marcinc, line 681: on all the Nodes (the VMs) -> on each of K8s cluster Nodes

forwarded to the service. -> forwarded to the ClusterIP service which is automatically created.

docs/reference/config-params.md
```diff
@@ -634,9 +634,13 @@ The `service` group contains configuration detail around Kubernetes services and
...
 
-Defines the type of Kubernetes service for a specific workload. See official K8s [documentation](https://kubernetes.io/docs/concepts/services-networking/service/). Kev will attempt to extract that information from the compose configuration.
+Defines the type of Kubernetes service for a specific workload. See official K8s [documentation](https://kubernetes.io/docs/concepts/services-networking/service/).
 
-The following heuristic determines the service type for application component:
+Although Kev provides a variety of types you can use, it only tries to extract two types of services from the compose configuration, namely `None` or `ClusterIP`.
+
+If you need a different type, please configure it manually. The different types are listed and explained below. 
+
+Here is the heuristic used to extract a `None` or `ClusterIP` type:
```
marcinc, line 643: a `None` or `ClusterIP` type -> a service type.

docs/reference/config-params.md
```diff
@@ -634,9 +634,13 @@ The `service` group contains configuration detail around Kubernetes services and
 
 ## kev.service.type
 
-Defines the type of Kubernetes service for a specific workload. See official K8s [documentation](https://kubernetes.io/docs/concepts/services-networking/service/). Kev will attempt to extract that information from the compose configuration.
+Defines the type of Kubernetes service for a specific workload. See official K8s [documentation](https://kubernetes.io/docs/concepts/services-networking/service/).
 
-The following heuristic determines the service type for application component:
+Although Kev provides a variety of types you can use, it only tries to extract two types of services from the compose configuration, namely `None` or `ClusterIP`.
+
+If you need a different type, please configure it manually. The different types are listed and explained below.
```
marcinc, line 641: Maybe worth adding a link to the official docs here too? https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types

ezodude: We have a link above `... See official K8s [documentation](https://kubernetes.io/docs/concepts/services-networking/service/).` ... I'll replace that url with the url suggested.

marcinc (approved):

PR #322, opened 2020-10-21
Init: always creates an implicit `dev` environment.

### LINKED ISSUE #317: Init: always creates an implicit `dev` environment.
## Problem

As a developer, I would like a sandbox environment so that I can experiment with builds and deployments.

## Solution

`kev init` always creates an implicit `dev` environment:
- This is created regardless of the `--environment` flags supplied by the user.
- If a user supplies `--environment` flag also named `dev` only one environment  will be created.

ezodude:
Resolves #317

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,13 @@
 ![License: Apache-2.0](https://img.shields.io/github/license/appvia/kev)
 
 
-_Kev_ helps developers port and iterate Docker Compose apps onto Kubernetes. It understands the Docker Compose application topology and prepares it for deployment in (multiple) target environments, with minimal user input.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -78,18 +78,19 @@ It makes use of,
 Creating the files below in your project directory:
 
 ```sh
-├── docker-compose.kev.stage.yaml       # stage Compose environment override file
+├── docker-compose.kev.stage.yaml       # dev sandbox Compose environment override file
```
marcinc, line 81: this should be `docker-compose.kev.dev.yaml` I guess?

marcinc (approved):
one small comment, LGO.

README.md
ezodude: Oops, will fix.

marcinc (approved):

Issue #361, opened 2021-01-06
Respect and support service extensions in environment overrides


ezodude:
A service config extension can be a simple way of layering custom behaviour to a specific environment.

However, Kev environment override files do not currently respect service config extensions. If an extension is defined, it is stripped out and never merged back into the source.

Kev environment override files should be extended to support service config extensions.

PR #362, opened 2021-01-07
Respect and support service extensions in environment overrides

ezodude:
Resolves: #361 

This PR implements the following extension strategy:

- Service config extensions can be added to the compose sources.
- Service config extensions can be added to the environment override files.
- Service config extensions will always be merged back to the sources config.
   - **Pre merge**: If an extension does not exist in the source ... **Post merge**: it will be added to the source.
   - **Pre merge**: If an extension exists in the source ... **Post merge**: it will be merged in the source, with the environment 
   override extension entries overriding the source's entries.
- Extensions are not reconciled.

This strategy allows us to add extensions to environment overrides, that will persist even if they don't exist in the original compose sources. And, merge in overrides if compose sources already contain the extension.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,12 +4,18 @@
 **/kev.yml
 **/docker-compose.kev.*.yaml
 k8s/*
+docker-compose.yaml
 !/pkg/kev/testdata/**/*.yaml
 !/examples/**/*.yaml
+skaffold.yaml
 
 ### IDEs ###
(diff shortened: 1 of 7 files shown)


marcinc (commented):
LG. Worth adding a spec for case where extensions are only present in the override?

marcinc (approved):

PR #374, opened 2021-02-02
Enables exclude list in render logic

### LINKED ISSUE #373: Support project service exclude list
Currently all compose project services will be included in render unless specifically excluded with `kev.component.enabled` flag. 

In some cases we might need to exclude a number of compose project services per given environment, which may be determined externally. 

The scope of this issue covers underlying elements to enabled such exclusion in the tool.

marcinc:
Resolves #373


ezodude (approved):
👍

PR #382, opened 2021-02-16
 Refactor init to reduce logic in CLI interface

ezodude:


pkg/kev/manifest.go
```diff
@@ -279,3 +280,17 @@ func (m *Manifest) getSourcesOverride() *composeOverride {
...
+
+func ManifestExistsForPath(manifestPath string) bool {
+	_, err := os.Stat(manifestPath)
+	return err == nil
+}
+
+func ensureFirstInit(wd string) error {
+	manifestPath := path.Join(wd, ManifestName)
+	if ManifestExistsForPath(manifestPath) {
+		err := fmt.Errorf("kev.yaml already exists at: %s", manifestPath)
```
marcinc, line 292: Could replace kev.yaml with `ManifestName` const, in case it ever changes.

marcinc (approved):

Issue #383, opened 2021-02-18
Derive environment override file name suffix from the ManifestName


marcinc:
Currently environment override files are named with a suffix that is hardcoded. 

Should top level ManifestName change this won't be reflected in other places in the codebase. 

Proposal: Allow for dynamic suffix in the environment override files to be extracted from currently set Manifest name.

PR #384, opened 2021-02-18
Derive environment override file name from kev manifest file name. 

marcinc:
Resolves: #383 

This PR modifies how the environment overrides are named. The suffix is now extracted from the ManifestName rather than being fixed.

### DIFF
--- a/pkg/kev/kev.go
+++ b/pkg/kev/kev.go
@@ -29,14 +29,14 @@ import (
 	"github.com/pkg/errors"
 )
 
-var (
-	// ManifestName main application manifest
-	ManifestName = "kev.yaml"
-)
-
 const (
 	// SandboxEnv is a default environment name
 	SandboxEnv = "dev"
+)
+
+var (
+	// ManifestFilename is a name of main application manifest file
+	ManifestFilename = "kev.yaml"
 )
 
 // InitProjectWithOptions initialises a kev project using provided options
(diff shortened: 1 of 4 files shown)


pkg/kev/kev.go
```diff
@@ -29,10 +29,14 @@ import (
...
 )
 
-const (
+var (
 	// ManifestName main application manifest
 	ManifestName = "kev.yaml"
-	SandboxEnv   = "dev"
+)
+
+const (
```
ezodude, line 37: Can we move the const declaration to the top of the file?

pkg/kev/kev.go
```diff
@@ -29,10 +29,14 @@ import (
 	"github.com/pkg/errors"
 )
 
-const (
+var (
 	// ManifestName main application manifest
 	ManifestName = "kev.yaml"
```
ezodude, line 34: How about if we change this to `ManifestFilename`?

Then `symbol` in `MintEnvironments` fun can simply be called `manifestName`?

pkg/kev/manifest.go
```diff
@@ -119,6 +119,8 @@ func (m *Manifest) CalculateSourcesBaseOverride(opts ...BaseOverrideOpts) (*Mani
 func (m *Manifest) MintEnvironments(candidates []string) *Manifest {
 	fileNameTemplate := m.GetEnvironmentFileNameTemplate()
 
+	symbol := strings.TrimSuffix(ManifestName, filepath.Ext(ManifestName))
```
ezodude, line 122: See prior comments regarding `ManifestName` and `symbol`. Additionally, this would be quite useful as a general function: `GetManifestName()`.

ezodude (commented):
Left some comments, but the general gist is correct.

ezodude (approved):
👍

PR #389, opened 2021-03-02
Only validate kev specific configuration labels

### LINKED ISSUE #388: Validate only kev specific configuration labels
Currently all configuration labels are validated which prevents user from specifying labels that should become annotations on generated objects. 

Solution: Only validate configuration labels starting with recognised label prefix. 

Labels that aren't related to kev configuration should be ignored and not validated. This will allow for treating them as annotations.

marcinc:
Resolves #388


ezodude (approved):
👍 (thanks for the walkthrough)

PR #390, opened 2021-03-04
Support dependent env vars

marcinc:
Resolves: #391 

Compose Environment variables which values are dependent on other environment variables must be ordered properly so that they can be expanded correctly. This PR allows for env variables with templated values that reference other environment variables, e.g.

```yaml
# docker-compose.yaml

environment:
  MY_VAR: {{FOO}}/{{BAR}}
  FOO: foo
  BAR: bar
```

In k8s manifest this will translate to:

```yaml
    spec:
      containers:
        - env:
            - name: FOO
              value: "foo"
            - name: BAR
              value: "bar"
            - name: MY_VAR
              value: "$(FOO)/$(BAR)"
```

### DIFF
--- a/docs/reference/config-params.md
+++ b/docs/reference/config-params.md
@@ -700,7 +700,7 @@ All traffic on the port you specify will be forwarded to the service allowing an
 
 Again, it is ideal for exposing a service or app to the internet under a single IP address.
 
-Practically, in non development environments, a LoadBalancer will be used to route traffic to an Ingress to expose multiple services under the same IP address and keep your costs down.     
+Practically, in non development environments, a LoadBalancer will be used to route traffic to an Ingress to expose multiple services under the same IP address and keep your costs down.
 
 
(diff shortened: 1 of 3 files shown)


ezodude (approved):
💯

docs/reference/config-params.md
```diff
@@ -842,6 +842,21 @@ services:
...
+services:
+  my-service:
+    labels:
+      ...
+    environment:
+      ENV_VAR_A: foo
+      ENV_VAR_B: bar
+      ENV_VAR_C: {{ENV_VAR_A}}/{{ENV_VAR_B}}  # referencing other dependent environment variables
+```
+
```
ezodude, line 859: This make a lot of sense.

In fact, we need this as a custom method for relying on ENV values, that will pick the value of another ENV value. Esp., that compose doesn't have a native mechanism to do this.

ezodude (approved):
👍

PR #412, opened 2021-03-18
kev-409 Extend json validation for probes

mangas:


pkg/kev/config/schema.go
```diff
@@ -67,6 +67,82 @@ var ServicesSchema = map[string]interface{}{
...
+						LabelWorkloadLivenessProbeType: map[string]interface{}{
+							"const": "command",
+						},
+					},
+					"required": []string{LabelWorkloadLivenessProbeCommand},
+				},
+				// This acts as a catch all when the required properties are not provided, should come in the end
+				// otherwise instead of matching that "command" type needs a properties the error says the value is not "none".
+				{
+					"properties": map[string]interface{}{
```
ezodude, line 101: @mangas @marcinc I'm curious about the following situation: 

... Changing the default `kev.workload.liveness-probe-type` from `command` to `none`
... On render

1 - Should we ignore the existing `kev.workload.liveness-probe-command` and render (without setting a probe)?

OR

2 - Should we error and instruct the user to remove `kev.workload.liveness-probe-command` as its no used/needed?

**Currently the code does (1)**

ezodude (commented):
Would appreciate feedback on comment.

pkg/kev/config/schema.go
mangas: I thought about this and the use case I came with is if I want to disable the problem I can just change to none. I don't think we should error on that since the extra parameter doesn't really matter. Otherwise I'd have to remove the whole thing and add it back. The other side of this argument is that we could comment. Also raises the question of whether or not we care about extra properties enough to error because of them? Does that make sense?

marcinc: That's a case I brought up before around a use case where one iterates on the app with liveness probe defined, then temporarily turns it off (e.g. when testing), to then turn it back on. 

The current behaviour does the (1) as you said above, however, `render` will respect it and won't generate the probe body. This also enables the user to turn the probe back on by setting type to `command` without having to redefine the command. 

The only slightly negative side effect of the use case above could be that when user decides to drop the probe entirely by setting its type to `none` AND at the same time forgets to remove the command itself, then it'll remain untouched in service configuration labels. But I think we can live with that.

pkg/kev/config/schema.go
```diff
@@ -67,6 +67,82 @@ var ServicesSchema = map[string]interface{}{
...
+						LabelWorkloadLivenessProbeType: map[string]interface{}{
+							"const": "command",
+						},
+					},
+					"required": []string{LabelWorkloadLivenessProbeCommand},
+				},
+				// This acts as a catch all when the required properties are not provided, should come in the end
+				// otherwise instead of matching that "command" type needs a properties the error says the value is not "none".
```
marcinc, line 99: slight rewording of this comment would be beneficial

pkg/kev/services.go
```diff
@@ -135,6 +135,23 @@ func (sc ServiceConfig) validate() error {
...
 
 	if !result.Valid() {
+		// Prioritise clear error messages.
+		for _, e := range result.Errors() {
+			if e.Type() == "required" {
+				return errors.New(e.Description())
+			}
+		}
+
+		for _, e := range result.Errors() {
```
marcinc, line 145: do we need to iterate through `result.Errors()` twice?

pkg/kev/services_test.go
```diff
@@ -48,8 +49,9 @@ var _ = Describe("ServiceConfig", func() {
 
 			It("success if the necessary labels are present", func() {
 				err := ServiceConfig{Labels: composego.Labels{
-					config.LabelWorkloadLivenessProbeType: kubernetes.ProbeTypeCommand.String(),
-					config.LabelWorkloadReplicas:          "1",
+					config.LabelWorkloadLivenessProbeType:    kubernetes.ProbeTypeCommand.String(),
```
marcinc, line 52: I guess a couple more specs covering other probe types (for both expected success and failure cases) would be good to have as well.

marcinc (commented):
Left a few comments. Should this PR also address the dependent config fields validations, based on selected probe type?

pkg/kev/config/schema.go
ezodude: This can get very confusing fast - even in other scenarios.

You can start collecting a lot of cruft.

For example this override works/renders - we had an `http` probe that got swapped to a `command` probe.

```yaml
...
  db-test:
    labels:
      kev.workload.liveness-probe-command: '["CMD", "echo", "Define healthcheck command for service db-test"]'
      kev.workload.liveness-probe-type: command
      kev.workload.readiness-probe-http-path: /status
      kev.workload.readiness-probe-http-port: "8080"
      kev.workload.replicas: "1"
...
```

I guess a **Warning** can help here ... (follow up PR?).

marcinc: I'd agree here. There is potential here for leftover buildup. Either a Warning, or straight cleanup at the cost of slight inconvenience on user's part - as this will require providing the details to re-enabled probe again which personally I don't think is a much of a problem, and this use case shouldn't happen too often I don't think.

pkg/kev/config/schema.go
mangas: do you have a specific suggestion besides fixing the typo?

pkg/kev/services.go
mangas: I mean we could do it in one at the cost of complexity, I think this is worth doing because it keeps the code simple. We want to prioritise any "required" errors which means we need to visit all errors. I can keep track "non required" and return it at the end if no "required" is found but I thought it would be a bit more confusing when reading.

pkg/kev/services_test.go
mangas: this is very covered throughout but I can add some more specific to the schema validation

pkg/kev/config/schema.go
mangas: https://github.com/appvia/kev/issues/414

pkg/kev/config/schema.go
marcinc: I think just a simple rephrase would help to make it clearer to follow.

pkg/kev/services.go
marcinc: OK, I was questioning the need for a second for loop above. You could just move the checks you are making here to the first for loop above. You check the types in expected sequence and return correct value anyway so no point in having two loop here.

pkg/kev/services_test.go
marcinc: Yeh, would be good to add. Thanks.

pkg/kev/services.go
mangas: I removed it, I think it's still kind of ok to understand. Let me know what you think

pkg/kev/config/schema.go
mangas: I removed the why and just left the what to do 😆

pkg/kev/services.go
```diff
@@ -135,24 +135,25 @@ func (sc ServiceConfig) validate() error {
...
 
-		for _, e := range result.Errors() {
 			// These errors are very cryptic and hurt usability.
 			if e.Type() == "number_one_of" || e.Type() == "number_any_of" || e.Type() == "number_all_of" {
 				continue
 			}
 
-			return errors.New(e.Description())
+			// this is a valid error if there isn't anything better
+			err = errors.New(e.Description())
```
marcinc, line 153: Don't you need to return here? The above will return the last error message - was it an intention?

pkg/kev/services.go
```diff
@@ -135,24 +135,25 @@ func (sc ServiceConfig) validate() error {
...
 		}
 
-		// If we don't find anything useful just go with whatever is available.
-		return errors.New(result.Errors()[0].Description())
+		return err
```
marcinc, line 156: and if you return above there is no need to return here and initialise err at the top of this funcion.

pkg/kev/services_test.go
```diff
@@ -141,6 +143,190 @@ var _ = Describe("ServiceConfig", func() {
...
+						serviceConfig.Labels[config.LabelWorkloadReadinessProbeType] = "wrong-value"
+
+						err := serviceConfig.validate()
+						Expect(err).To(HaveOccurred())
+						Expect(err.Error()).To(ContainSubstring(config.LabelWorkloadReadinessProbeType))
+						Expect(err.Error()).To(ContainSubstring("none"))
+					})
+
+					It("succeeds if set to none", func() {
+						serviceConfig.Labels[config.LabelWorkloadLivenessProbeType] = kubernetes.ProbeTypeNone.String()
```
marcinc, line 301: I guess you meant `config.LabelWorkloadLivenessProbeType` -> `config.LabelWorkloadReadinessProbeType`

pkg/kev/services_test.go
```diff
@@ -141,6 +143,190 @@ var _ = Describe("ServiceConfig", func() {
...
+					It("fails when command probe has missing properties", func() {
+						err := serviceConfig.validate()
+						Expect(err).To(HaveOccurred())
+						Expect(err.Error()).To(ContainSubstring("required"))
+						Expect(err.Error()).To(ContainSubstring(config.LabelWorkloadLivenessProbeCommand))
+					})
+				})
+			})
+
+			Context("liveness probe type enum validation", func() {
```
marcinc, line 285: if you context covers both liveness and readiness probe specs then perhaps rename it to `probe type enum validation` instead

pkg/kev/services_test.go
```diff
@@ -141,6 +143,190 @@ var _ = Describe("ServiceConfig", func() {
...
+				BeforeEach(func() {
+					serviceConfig.Labels[config.LabelWorkloadLivenessProbeType] = kubernetes.ProbeTypeNone.String()
```
marcinc, line 287: I think it'd be better to move this `BeforeEach` block to relevant context, for clarity, or remove entirely if not used.

marcinc (commented):
A few more little comments

pkg/kev/services_test.go
mangas: added

pkg/kev/services.go
mangas: this is what makes this thing weird, if you get to this line you still have no guarantee the one of the next iterations isn't a "required" error. so we want to make sure we keep one "good enough" error around in case there ins't a required. Does that make sense?

pkg/kev/services.go
mangas: responded in the previous comment

pkg/kev/services_test.go
mangas: yes, forgot to change this one, thanks !

pkg/kev/services_test.go
mangas: This should go under readiness

pkg/kev/services_test.go
mangas: this can actually be removed because it's set on the BeforeEach, good catch

pkg/kev/services_test.go
```diff
@@ -141,6 +143,189 @@ var _ = Describe("ServiceConfig", func() {
...
+						serviceConfig.Labels[config.LabelWorkloadLivenessProbeTCPPort] = "8080"
+
+						err := serviceConfig.validate()
+						Expect(err).To(Succeed())
+					})
+				})
+			})
+
+			Context("command probe validation", func() {
+				Context("rediness probess", func() {
```
ezodude, line 257: `rediness probess` -> `readiness probe`?

Or just `readiness`? maybe the context below can also be just `liveness`?

ezodude (commented):
All good - minor typos and such...

pkg/kev/services.go
ezodude: Feels there should be a better way to address the above.

- Perhaps first finding an error with `e.Type() == "required"` ... maybe using `go-funk` -> `funk.Find(...)
- And then finding any error that is NOT `e.Type() == "number_one_of" || e.Type() == "number_any_of" || e.Type() == "number_all_of"` and returning.

See https://github.com/thoas/go-funk

mangas: Find will do the exact same thing my original code did which is basically iterate and check if there is a match to the predicate. And then that's exactly what I was doing, not sure there is a point in introducing a library to avoid writing a for-loop?

pkg/kev/services.go
```diff
@@ -134,9 +134,52 @@ func (sc ServiceConfig) validate() error {
...
+
+	// If we don't find anything useful just go with whatever is available.
+	return errors.New(result.Errors()[0].Description())
+}
+
+func excludeTypes(ts ...string) func(gojsonschema.ResultError) bool {
+	return func(re gojsonschema.ResultError) bool {
+		for _, t := range ts {
+			if t == re.Type() {
+				return true
```
marcinc, line 159: as you exclude I guess this should return false, and true by default otherwise?

ezodude: I agree with @marcinc here.

ezodude: Otherwise, it makes sense. 👌

As a final nice to have, it would be great if we can move the schema logic into a separate file if possible.

mangas: yep, this is why the build was failing, it's fixed now. Do you think this is more clear/easier to understand without a lot of context?

mangas: also I agree with this not being the right place for this but is there a good place? Is it going to be re-used somewhere else? I wonder if this would make sense as a contribution to the upstream repo

ezodude (approved):

PR #415, opened 2021-03-19
Extract out the terminal UI lib into its own repo.

### LINKED ISSUE #413: Extract out the terminal UI lib into its own repo.

ezodude:
Resolves #413 

Replaces use of the `terminal` pkg with the new [komando](https://github.com/appvia/komando) pkg.


pkg/kev/init.go
```diff
@@ -22,7 +22,7 @@ import (
 	"path/filepath"
 
 	"github.com/appvia/kev/pkg/kev/config"
-	"github.com/appvia/kev/pkg/kev/terminal"
+	"github.com/appvia/komando"
```
marcinc, line 25: Tiny one. Could alias with something shorter eg `kmd` or similar perhaps?

marcinc (commented):
One tiny comment. LG Otherwise.

marcinc (approved):
👍

PR #426, opened 2021-03-23
update examples

mangas:


examples/wordpress-mysql/docker-compose.yaml
```diff
@@ -2,7 +2,7 @@ version: '3.7'
 services:
 
   db:
-    image: mysql:5.7
+    image: mysql:8.0
```
marcinc, line 5: tiny one, would we good to check whether it all still works with 8.0 - don't expect any problems here btw.

mangas: I had to bump it because it wasn't working before but it is now

ezodude: Yup, we standardised on 5.7 as newer versions weren't working at the time. But maybe that's fixed now :)

docs/reference/config-params.md
```diff
@@ -380,26 +380,26 @@ services:
...
+If compose file(s) specifies the `healthcheck.disable` attribute key in a service config it set the probe type to `none`.
+Otherwise it'll default to `command` (liveness probe active!)
 
-### Default: `false`
+### Default: `command`
 
-### Possible options: Bool
+### Possible options: none, command, http, tcp.
 
 > kev.workload.liveness-probe-disabled:
```
ezodude, line 396: `kev.workload.liveness-probe-disabled:` -> `kev.workload.liveness-probe-type:`

docs/reference/config-params.md
```diff
@@ -380,26 +380,26 @@ services:
...
 ```yaml
 version: 3.7
 services:
   my-service:
     labels:
-      kev.workload.liveness-probe-disabled: true
+      kev.workload.liveness-probe-type: http
```
ezodude, line 402: For simplicity, should this be `....-type: none`? I ask, because we're not showing the expected related `http` probe labels.

docs/reference/config-example.md
```diff
@@ -25,12 +25,32 @@ services:                                                # compose services sect
...
+      kev.workload.liveness-probe-type: "command"        # Default: command. Possible options: none | command | http | tcp.
+      ## Liveness probe
+      ### Command
+      kev.workload.liveness-probe-command: "echo 'n/a'"  # Liveness probe command to run. Only used when using command probe type.
+      ### HTTP
+      kev.workload.liveness-probe-http-port: 8080        # HTTP Liveness probe port. Only used when using an http probe type.
+      kev.workload.liveness-probe-http-path: /status     # HTTP Liveness probe path. Only used when using an http probe type.
+      ### TCP
+      kev.workload.liveness-probe-tcp-port: 8080         # TCP Liveness probe port. Only used when using an tcp probe type.
+      kev.workload.liveness-probe-initial-delay: 1m0s    # Default: 1m. How long to wait before initial probe run. Shared by all liveness probe types.
```
marcinc, line 37: perhaps move the last four lines up to the liveness probe as they are common / shared across all liveness probe types. Similar comment for the readiness probe below.

docs/reference/config-params.md
```diff
@@ -518,21 +569,21 @@ services:
...
 
-> kev.workload.readiness-probe-disabled:
+> kev.workload.readiness-probe-type:
 ```yaml
 version: 3.7
 services:
   my-service:
     labels:
-      kev.workload.readiness-probe-disabled: true
+      kev.workload.readiness-probe-type: http
```
ezodude, line 586: Same as https://github.com/appvia/kev/pull/426/files#diff-12e7c8a5e801707eb612dbb78efb11ccaf1c0f264feca5f5653a90964baf75f6R402 ?

ezodude (commented):
Really enjoyed the fixes to parsing the port number. 🌮⚡️

Left some documentation comments.

docs/reference/config-params.md
mangas: sure, I just thought having a "read example" could be good instead of copying and then changing it since realistically no one is going to add none 😛  but yes I get the point

docs/reference/config-example.md
mangas: but aren't these secondary? as in, they even have defaults and can be omitted. What about adding a ### Common section?

marcinc: yes please. commend would do.. something to tell that these are common for all liveness probe types.

marcinc (approved):
LGTM 👍

PR #432, opened 2021-03-29
add extension support

mangas:


examples/node-app/docker-compose.yaml
```diff
@@ -6,3 +6,11 @@ services:
       context: .
     ports:
       - "8080:8080"
+    x-k8s:
+      liveness-probe:
+        disabled: false
```
marcinc, line 11: I'm not sure you'd noticed but I changed the spec for it slightly in the related issue. Now that we support multiple types it'd make sense to reflect it in this block as well.

mangas: this is jsut for testing, still a draft PR. I will remove this stuff from here or update it once it's supported

mangas: this is also still the old spec :) check here: https://github.com/appvia/kev/pull/432/files#diff-ad020f19dc8b4d3bd06cec983ef7d65389a25615a114919aa08bb865b83d696cR18 <- this should be the new spec

pkg/kev/config/extensions.go
```diff
@@ -0,0 +1,143 @@
...
+	k8s := DefaultK8SServiceConfig()
+
+	if err := mergo.Merge(&extensions.K8S, k8s); err != nil {
+		return K8SServiceConfig{}, err
+	}
+
+	return extensions.K8S, nil
+}
+
+// TODO: hyphenated words suck, should we do camelCase?
```
marcinc, line 68: Yes, I think it makes sense to switch to camelCase in extensions. Ideally let's do it as part of this initial PR please.

pkg/kev/config/extensions.go
```diff
@@ -0,0 +1,143 @@
...
+	ProbeConfig `yaml:",inline"`
+}
+
+func DefaultLivenessProbe() LivenessProbe {
+	id, _ := time.ParseDuration(DefaultProbeInitialDelay)
+	p, _ := time.ParseDuration(DefaultProbeInterval)
+
+	return LivenessProbe{
+		Type: ProbeTypeExec.String(),
+		ProbeConfig: ProbeConfig{
```
marcinc, line 88: `ProbeConfig` missing timeout option - also applicable to the ReadinessProbe below. 

It'd be great to have this (liveness/readiness probe) scoped portion of extension based config work as complete as we possibly can to cover the current state of play with the probes. This way we'll spend less time context switching between different parts in the future PRs. 

Transition of remaining configuration will be then scoped around their respective options in the follow up PR which will be a nice way of braking that work down in stages.

pkg/kev/config/extensions_test.go
```diff
@@ -0,0 +1,117 @@
...
+					parsedCfg, err = config.K8SServiceCfgFromMap(extensions)
+					Expect(err).To(HaveOccurred())
+					Expect(err.Error()).To(Equal("Workload.Replicas is required"))
+				})
+			})
+		})
+	})
+
+	Describe("Marshalling", func() {
+		It("Invariance", func() {
```
marcinc, line 87: minor: would be good to give more description as to what this case is actually testing.

pkg/kev/config/extensions_test.go
```diff
@@ -0,0 +1,117 @@
...
+			Expect(err).To(Succeed())
+		})
+
+		Context("works with defaults", func() {
+			BeforeEach(func() {
+				k8s.Workload.Replicas = 10
+				k8s.Workload.LivenessProbe.Type = config.ProbeTypeNone.String()
+			})
+
+			It("Works with defaults", func() {
```
marcinc, line 43: would be great to differentiate or put some more detail in that case description. Currently it's identical to the context description...

pkg/kev/config/extensions.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+type ExtensionRoot struct {
+	K8S K8SServiceConfig `yaml:"x-k8s"`
+}
+
+type K8SServiceConfig struct {
+	Enabled   bool      `yaml:"enabled"`
+	Workload  Workload  `yaml:"workload" validate:"required"`
+	Component Component `yaml:"component"`
```
marcinc, line 21: `Component` is basically redundant here as we already have `Enabled` field above covering for what we currently support at a [component level](https://github.com/appvia/kev/blob/master/docs/reference/config-params.md#kevcomponentenabled).

Also, [K8s service part](https://github.com/appvia/kev/blob/master/docs/reference/config-params.md#-service) of configuration isn't cover yet, which is fine, but maybe drop it in there as a placeholder `Service` field with a todo action?

pkg/kev/config/extensions.go
```diff
@@ -0,0 +1,143 @@
...
+
+type TCPProbe struct {
+	Port int `yaml:"port"`
+}
+
+type ExecProbe struct {
+	Command string `yaml:"command"`
+}
+
+type Component struct {
```
marcinc, line 142: I don't think we need that give the comment above.

marcinc (commented):
A few comments.

pkg/kev/config/extensions_test.go
```diff
@@ -0,0 +1,117 @@
...
+		BeforeEach(func() {
+			k8s = config.K8SServiceConfig{}
+		})
+
+		JustBeforeEach(func() {
+			var buf bytes.Buffer
+			err := yaml.NewEncoder(&buf).Encode(map[string]interface{}{
+				"x-k8s": k8s,
+			})
+			Expect(err).To(Succeed())
```
ezodude, line 31: Canonically we've been using `ToNot(HaveOccurred())` in our tests.

From the [Gomega docs](https://onsi.github.io/gomega/#handling-errors) `Succeed()` asserts a function has returned without an error and is ideally used as:
```go
Expect(DoSomethingHard()).Should(Equal("foo"))
```

So IMHO we should either stick with `ToNot(Have...` or use `Succeed()` as set in the Gomega docs - but ideally not mix the two.

pkg/kev/config/extensions_test.go
```diff
@@ -0,0 +1,117 @@
...
+			Expect(expected).To(BeEquivalentTo(actual))
+		})
+
+		It("marshals invalid probetype as empty string", func() {
+			expected := config.DefaultLivenessProbe()
+			expected.Type = config.ProbeType("asd").String()
+
+			var buf bytes.Buffer
+			err := yaml.NewEncoder(&buf).Encode(expected)
+			Expect(err).To(Succeed())
```
ezodude, line 107: See comment above regarding `Succeed()`

pkg/kev/config/extensions_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+			var actual config.LivenessProbe
+			err = yaml.NewDecoder(&buf).Decode(&actual)
+			Expect(err).To(Succeed())
```
ezodude, line 111: See comment above regarding `Succeed()`

pkg/kev/config/types.go
```diff
@@ -0,0 +1,52 @@
+package config
```
ezodude, line 1: Can we rename this file? - we normally reserve `types.go` to contain actual types used in a pkg.

ezodude (commented):
Great work! - Left some comments.

pkg/kev/config/extensions.go
```diff
@@ -0,0 +1,162 @@
...
+// LivenessProbe holds all the settings for the same k8s probe.
+type LivenessProbe struct {
+	// TODO: find a decent way of using ProbeType here that validates the content of the string
+	Type        string `yaml:"type" validate:"required"`
+	ProbeConfig `yaml:",inline"`
+}
+
+// DefaultLivenessProbe creates a default liveness probe. Defaults to exec.
+func DefaultLivenessProbe() LivenessProbe {
+	id, _ := time.ParseDuration(DefaultProbeInitialDelay)
```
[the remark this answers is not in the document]
mangas: yep, it's a good point

pkg/kev/config/extensions.go
```diff
@@ -0,0 +1,162 @@
...
+	p, _ := time.ParseDuration(DefaultProbeInterval)
+	t, _ := time.ParseDuration(DefaultProbeTimeout)
```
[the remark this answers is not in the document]
mangas: done

pkg/kev/config/extensions_test.go
mangas: I blame them for having at least 8 different ways of expressing the same thing!

pkg/kev/config/types.go
mangas: moved

ezodude (approved):
🚀

PR #437, opened 2021-03-30
updated the service name in inital docker compose file due to configu…

Alhinde93:
…ration conflicts in latest wordpress image


Alhinde93:
When running through the getting started documentation, after completing the initial Kubernetes deployment, i was getting database connection error when port forwarding the service. After looking at the Wordpress configuration, there were two pieces of configuration present. This is due to the default Kubernetes env variables being named after the service.
The latest image will check to see if there are any environment variables present with the Wordpress prefix and generate a new config file. 
config [here](https://github.com/docker-library/wordpress/blob/master/latest/php8.0/apache/docker-entrypoint.sh#L70-L97  ) 
I have renamed the service to avoid this, can you please review?

myjaffoo (approved):

PR #439, opened 2021-03-31
add extensions to init

### LINKED ISSUE #438: wire K8SConfig to be added on init
kev init should dump the defined extensions or default if none was provided on the overlay files

Part of #443.

mangas:
Resolves #438


ezodude (commented):
I just did a quick review - makes sense.

Love that you renamed to `K8SConfiguration` 🚀 ... I'm sure @marcinc will give it a thorough once over 🙂.

pkg/kev/init_test.go
```diff
@@ -216,11 +217,31 @@ services:
       kev.workload.liveness-probe-command: '["CMD", "echo", "Define healthcheck command for service db"]'
       kev.workload.liveness-probe-type: exec
       kev.workload.replicas: "1"
+    x-k8s:
+      enabled: true
+      workload:
+        type: Deployment
+        replicas: 1
+        livenessProbe:
+          type: exec
```
marcinc, line 226: I appreciate it may not be in scope of this PR, but something we should consider.

Instead of specifying the probe type we could just use it as a key with nested configuration specific to selected probe type, see example in https://github.com/appvia/kev/issues/443.

```yaml
livenessProbe:
  exec:
    command: "your command"
...

readinessProbe:
  http:
    port: 8080
    path: /
...

readinessProbe:
  tcp:
    port: 8080
...
```

The only challenge with the above is a `none` probe type... but perhaps we could work on assumption that if `none` type was intended, then the probe shouldn't be present in the configuration at all?

pkg/kev/sources.go
```diff
@@ -33,7 +34,34 @@ func (s *Sources) CalculateBaseOverride(opts ...BaseOverrideOpts) error {
...
+			Name:       svc.Name,
+			Labels:     map[string]string{},
+			Extensions: svc.Extensions,
+		}
+
+		setDefaultLabels(&target)
+		extractVolumesLabels(ready, s.override)
+		extractLabels(svc, &target)
+
+		k8sConf, err := config.K8SCfgFromMap(svc.Extensions)
```
marcinc, line 52: We now bite into the extensions so would be good to update the function description a little bit, to reflect that perhaps?

NOTE: I review on assumption that this change is backward compatible with labels driven config.

mangas: updated

pkg/kev/init_test.go
mangas: I'd rather table this for now and discuss this as a separate issue. This causes a bit of problems because many of them can be filled and so we end up with more errors and making it really hard for someone to switch between types because they would need to remove all the alternatives or risk them being used on the k8s config. Not sure if that produces an error or not. Either case I think this is better as a ticket to be discussed, what do you think?

marcinc (approved):

PR #450, opened 2021-04-07
Revisit skaffold profiles

### LINKED ISSUE #411: Ensure that Skaffold profiles use correct push strategy for containers built locally
There is currently an issue around Skaffold profiles whereby a profile's push override forces artefacts to be always pushed to the external registry. 

The above scenario is not always desirable. Specifically when developing against local Kubernetes cluster such as kind, minikube or docker-desktop. 

This issues could be addressed by changing profiles makeup or splitting them up so that profiles can be stacked together efficiently. NOTE that multiple profiles can be applied to `skaffold` to achieve a desired effect.

`push` scenarios to cover:

1) local build with push disabled
- used when images are built and pushed to local docker daemon for consumption in local K8s cluster.

2) local build with push enabled
- used when iterating on the application stack locally but deployed to any remote K8s cluster.

There are already CI related profiles that could be potentially repurposed too.

marcinc:
Resolves #411 

This PR introduces the following changes:

* Removes unnecessary local profile for `minikube` and `docker-desktop` as these are handled natively.
* Removes `Push` option in Build phase of each environment profile - local profiles will never push (unless overridden), and when remote contexts are in use then push will happen automatically (unless overridden).  
* Simplifies naming convention for the additional profiles.
* Changes build artefact discovery algorithm - native Skaffold Analysis, followed by discovery from base docker-compose file.
* Adds `UpdateSkaffoldBuildArtifacts` method to enabled automatic build artefacts list update in `skaffold.yaml` - this happens at `render` time given the project was initialised with skaffold support.
* Moves "Build" part of the pipeline from individual profiles to the top level (global) pipeline for better readability.
* Minor Skaffold profiles refactoring changes.

### DIFF
--- a/pkg/kev/skaffold.go
+++ b/pkg/kev/skaffold.go
@@ -64,9 +64,6 @@ const (
 	// SkaffoldFileName is a file name of skaffold manifest
 	SkaffoldFileName = "skaffold.yaml"
 
-	// ProfileNamePrefix is a prefix to the added skaffold aprofile
-	ProfileNamePrefix = "kev-"
-
 	// EnvProfileNameSuffix is a suffix added to environment specific profile name
 	EnvProfileNameSuffix = "-env"
 
@@ -256,10 +253,10 @@ func (s *SkaffoldManifest) SetProfiles(envs []string) {
 // AdditionalProfiles adds additional Skaffold profiles
 func (s *SkaffoldManifest) AdditionalProfiles() {
 
-	if !s.profileNameExist(ProfileNamePrefix + "ci-local-build-no-push") {
+	if !s.profileNameExist("ci-local-build-no-push") {
 		// Helper profile for use in CI pipeline
 		s.Profiles = append(s.Profiles, latest.Profile{
-			Name: ProfileNamePrefix + "ci-local-build-no-push",
+			Name: "ci-local-build-no-push",
 			Pipeline: latest.Pipeline{
 				Build: latest.BuildConfig{
 					BuildType: latest.BuildType{
@@ -274,10 +271,10 @@ func (s *SkaffoldManifest) AdditionalProfiles() {
 		})
 	}
 
-	if !s.profileNameExist(ProfileNamePrefix + "ci-local-build-and-push") {
+	if !s.profileNameExist("ci-local-build-and-push") {
 		// Helper profile for use in CI pipeline
 		s.Profiles = append(s.Profiles, latest.Profile{
-			Name: ProfileNamePrefix + "ci-local-build-and-push",
+			Name: "ci-local-build-and-push",
 			Pipeline: latest.Pipeline{
 				Build: latest.BuildConfig{
 					BuildType: latest.BuildType{
--- a/pkg/kev/skaffold.go
+++ b/pkg/kev/skaffold.go
@@ -106,9 +106,9 @@ func LoadSkaffoldManifest(path string) (*SkaffoldManifest, error) {
 	return s, yaml.Unmarshal(data, &s)
 }
 
-// AddProfiles injects kev profiles to existing Skaffold manifest
+// InjectProfiles injects kev profiles to existing Skaffold manifest
 // Note, if profile name already exists in the skaffold manifest then profile won't be added
-func AddProfiles(path string, envs []string, includeAdditional bool) (*SkaffoldManifest, error) {
+func InjectProfiles(path string, envs []string, includeAdditional bool) (*SkaffoldManifest, error) {
 	skaffold, err := LoadSkaffoldManifest(path)
 	if err != nil {
 		return nil, err
@@ -638,7 +638,7 @@ Once you have skaffold.yaml in your project, make sure that Kev references it by
 	}
 
 	// Reconcile skaffold config and add potentially missing profiles before starting dev loop
-	reconciledSkaffoldConfig, err := AddProfiles(configPath, manifest.GetEnvironmentsNames(), true)
+	reconciledSkaffoldConfig, err := InjectProfiles(configPath, manifest.GetEnvironmentsNames(), true)
 	if err != nil {
 		return "", nil, errors.Wrap(err, "Couldn't reconcile Skaffold config - required profiles haven't been added.")
 	}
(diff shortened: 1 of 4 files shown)


pkg/kev/skaffold.go
```diff
@@ -109,22 +106,78 @@ func LoadSkaffoldManifest(path string) (*SkaffoldManifest, error) {
 	return s, yaml.Unmarshal(data, &s)
 }
 
-// AddProfiles injects kev profiles to existing Skaffold manifest
+// InjectProfiles injects kev profiles to existing Skaffold manifest
```
mangas, line 109: Not sure abou Add vs Inject, feels like They convey the same message and Add seems to be more used overall 🤔

pkg/kev/skaffold.go
```diff
@@ -109,22 +106,78 @@ func LoadSkaffoldManifest(path string) (*SkaffoldManifest, error) {
...
+		skaffold.SetAdditionalProfiles()
 	}
 
 	return skaffold, nil
 }
 
+// UpdateSkaffoldBuildArtifacts updates skaffold build artefacts with freshly discovered list of images and contexts.
+// Note, it'll persist updated build artefacts in the skaffold.yaml file only when change in build artefacts was detected.
+// Important: The last discovered images and contexts will be persisted (if changed)!
+func UpdateSkaffoldBuildArtifacts(path string, project *ComposeProject) error {
```
mangas, line 128: I wonder if the this skaffold specific stuff should go on a package which encapsulates all the skaffold specifics

pkg/kev/skaffold.go
```diff
@@ -251,47 +304,13 @@ func (s *SkaffoldManifest) SetProfiles(envs []string) {
...
-					KubeContext: "docker-desktop",
-				},
-			},
-		})
-	}
+// SetAdditionalProfiles adds additional Skaffold profiles
+func (s *SkaffoldManifest) SetAdditionalProfiles() {
 
-	if !s.profileNameExist(ProfileNamePrefix + "ci-build-no-push") {
+	if !s.profileNameExist("ci-local-build-no-push") {
```
mangas, line 310: I think we could benefit from making this a method of SkaffoldManifest as in: 

```
func (s *SkaffoldManifest) TryAddProfile(p latest.Profile) // or whatever the write type of profile is, I didn't check if there is an interface for this  
```

pkg/kev/skaffold.go
```diff
@@ -353,46 +370,55 @@ func (s *SkaffoldManifest) SetBuildArtifacts(analysis *Analysis, project *Compos
...
-				buildArtifacts[context] = image
-				break
-			} else {
-				buildArtifacts[context] = svcNameFromContext
+		if len(analysis.Images) > 0 {
+			// Check whether images detected by Analysis contain service image name derived from the
+			// context as that's the best we can do in order to match a service to a corresponding
+			// docker registry image.
+			// If no docker registry image was detected then we use service name as docker image name.
+			re := regexp.MustCompile(fmt.Sprintf(`.*\/%s`, svcImageNameFromContext))
```
mangas, line 396: Is this faster than [HasSuffix](https://golang.org/pkg/strings/#HasSuffix)?  Also if we're doing compilation in line, it would prolly be a better idea to return an error instead of panic, I see this was not changed but I'm thinking it could be cleaned up

pkg/kev/skaffold.go
```diff
@@ -353,46 +370,55 @@ func (s *SkaffoldManifest) SetBuildArtifacts(analysis *Analysis, project *Compos
...
+
+			for _, image := range analysis.Images {
+				if found := re.FindAllStringSubmatchIndex(image, -1); found != nil {
```
mangas, line 399: nitpick: I think the following would read better(unless I'm missing something from what the code does)  

```
buildArtifacts[context] = svcImageNameFromContext
if strings.HasSuffix(image, svcImageNameFromContext) {
     buildArtifacts[context] = image
    break
}
```

pkg/kev/skaffold.go
marcinc: May become part of future PR.

pkg/kev/skaffold.go
marcinc: Agree, they both mean the same in this context. Changed to `Inject` to indicate that this is something we normally only do to an already existing Skaffold manifest, i.e. we inject additional external profiles to it which weren't there initially. The surface area of this func is small so will leave as is for now.

pkg/kev/skaffold.go
marcinc: Agreed, number of images detected shouldn't be significant so checking suffix will be cleaner from legibility standpoint.

pkg/kev/skaffold.go
```diff
@@ -353,46 +370,53 @@ func (s *SkaffoldManifest) SetBuildArtifacts(analysis *Analysis, project *Compos
...
-		// then we use service name as docker image name.
-		re := regexp.MustCompile(fmt.Sprintf(`.*\/%s`, svcNameFromContext))
-
-		for _, image := range analysis.Images {
-			if found := re.FindAllStringSubmatchIndex(image, -1); found != nil {
-				buildArtifacts[context] = image
-				break
-			} else {
-				buildArtifacts[context] = svcNameFromContext
+		if len(analysis.Images) > 0 {
```
mangas, line 391: As per my previous comment, this whole block can be reduced to the following:

```
buildArtifacts[context] = svcImageNameFromContext
for _, image := range analysis.Images {
 if strings.HasSuffix(image, svcImageNameFromContext) {
    buildArtifacts[context] = image
   break
 }
}

```

mangas (dismissed):
left comments

pkg/kev/skaffold.go
marcinc: If you look at the code we differentiate between two cases: 1) images have been detected by Analysis - this covers the case you mentioned, and 2) situation where there is no images detected and we need to fall back to what's available.

ezodude (approved):
I think @mangas already addressed the nitpick-y code bits. 👨‍💻👍

Generally, makes more sense. Esp. using the global profile and relying on kubecontext to determine the push strategy and rebuilding artefacts during render. 💯 

Quite a hefty piece of work. IMHO we should try break some of this down a bit in the future. Maybe 2 PRs?
- Push/profile re-work.
- Handling of build artefacts.

marcinc:
@ezodude Yeh, extract / refactor to follow in the ongoing PRs.

mangas (approved):

Issue #449, opened 2021-04-07
ensure overlay gets updated and merged with base during render


mangas:
ensure overlay gets updated and merged with base during render

PR #451, opened 2021-04-07
kev-449 overlay validation

mangas:
closes https://github.com/appvia/kev/issues/449


pkg/kev/manifest.go
```diff
@@ -189,6 +199,17 @@ func (m *Manifest) ReconcileConfig(envs ...string) (*Manifest, error) {
 	return m, nil
 }
 
+func validateExtensions(services Services) error {
+	for _, s := range services {
+		_, err := config.K8SCfgFromMap(s.Extensions)
+		if err != nil {
+			return errors.Wrapf(err, "unable to validation extensions for service %s", s.Name)
```
marcinc, line 206: How about `K8s configuration validation failed for service %s` instead?

pkg/kev/sources.go
```diff
@@ -86,6 +91,8 @@ func withEnvVars(s *Sources, origin *ComposeProject) error {
 			Name:        svc.Name,
 			Labels:      svc.Labels,
 			Environment: originSvc.Environment,
+			Extensions:  svc.Extensions,
+			K8SConfig:   svc.K8SConfig,
```
marcinc, line 95: Do we need K8sConfig here given we already track Extensions?

marcinc (commented):
A first pass with few comments & questions

pkg/kev/manifest.go
mangas: ahah yep I actually had fixed this 😛

pkg/kev/sources.go
mangas: will remove it now, I don't think we want to cache anything

pkg/kev/config/extensions_test.go
```diff
@@ -54,19 +54,60 @@ var _ = Describe("Extentions", func() {
...
+			})
+			Context("with RequirePresent option", func() {
+				It("unless the RequirePresnet option is specified", func() {
+					parsedCfg, err = config.K8SCfgFromMap(map[string]interface{}{}, config.RequirePresent())
+					Expect(err).To(HaveOccurred())
+				})
 			})
 		})
 
 		Describe("fails on mandatory fields", func() {
```
marcinc, line 77: Small rephrase suggestion `fails on mandatory fields` -> `validations`

pkg/kev/config/extensions_test.go
```diff
@@ -124,4 +165,43 @@ var _ = Describe("Extentions", func() {
...
+			expected := k8sBase
+			expected.Workload.Replicas = 10
+			expected.Workload.LivenessProbe.Type = config.ProbeTypeNone.String()
+
+			result, err := k8sBase.Merge(k8sTarget)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).To(BeEquivalentTo(expected))
+		})
+
+		When("results in default config are empty/invalid", func() {
```
marcinc, line 184: The bodies of two cases in this context are practically the same.. I think it'd make sense to turn the below into two separate context and set the initial condition for the test cases within those contexts. 

For the first one, how about the following format?

```go
When("k8s extension map is empty", func() {
    BeforeEach(func() {
        extensions := make(map[string]interface{})
        parsedConf, err := config.K8SCfgFromMap(extensions)
        Expect(err).NotTo(HaveOccurred())
    })

    It("doesn't return an error and resultant configuration matches the default configuration", func() {
        result, err := config.DefaultK8SConfig().Merge(parsedConf)
        Expect(err).NotTo(HaveOccurred())
        Expect(result).To(BeEquivalentTo(config.DefaultK8SConfig()))
    })
})
```

and the case after that could follow the same format for consistency.

pkg/kev/kev_reconcile_test.go
```diff
@@ -51,7 +51,7 @@ var _ = Describe("Reconcile", func() {
 		r := kev.NewRenderRunner(workingDir, kev.WithUI(kmd.NoOpUI()))
 		r.LoadProject()
 		manifest, mErr = r.Manifest().ReconcileConfig()
-		Expect(mErr).NotTo(HaveOccurred())
+		Expect(mErr).NotTo(HaveOccurred(), workingDir)
```
marcinc, line 54: why do we need `workingDir` here?

pkg/kev/testdata/reconcile-env-var-override/docker-compose.kev.dev.yaml
```diff
@@ -1,6 +1,17 @@
 version: "3.9"
 services:
   db:
+    x-k8s:
+      workload:
+        replicas: 1
+        livenessProbe: 
+          type: exec
+          exec:
+            command: ["CMD", "echo", "Define healthcheck command for service db"]
```
marcinc, line 10: Ideally we should move toward probe command in a format accepted by the K8s and not use docker compose healthcheck test format, i.e. 

```yaml
exec:
  command:
    - echo
    - Define healthcheck...
```

When we parse docker compose healthcheck command it must be defined with one of `NONE`, `CMD`, `CMD-SHELL` - as per https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck. When we transform that healthcheck into something K8s understands we drop that first element, see: https://github.com/appvia/kev/blob/master/pkg/kev/converter/kubernetes/project_service.go#L675

marcinc (commented):
Left a few more comments.

pkg/kev/config/extensions.go
```diff
@@ -30,13 +73,33 @@ func DefaultK8SConfig() K8SConfiguration {
...
 
+type k8sConfigOptions struct {
+	requirePresent bool
+}
+
+// K8SCfgOption will modify parsing behaviour of the x-k8s extension.
+type K8SCfgOption func(*k8sConfigOptions)
+
+// RequirePresent will ensure that x-k8s is present and that it is validated.
+func RequirePresent() K8SCfgOption {
```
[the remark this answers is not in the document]
mangas: Along those lines I think `RequireExtensions` is not bloated and reads similarly, do you agree?

pkg/kev/kev_reconcile_test.go
mangas: because otherwise we don't know which workingDir is failing it just says that something is failing in this line, this way if it fails here you can easily pinpoint the test since this property is mutated in every test pretty much

pkg/kev/testdata/reconcile-env-var-override/docker-compose.kev.dev.yaml
mangas: updated on all files

pkg/kev/config/extensions_test.go
```diff
@@ -124,4 +171,43 @@ var _ = Describe("Extentions", func() {
...
+			expected.Workload.LivenessProbe.Type = config.ProbeTypeNone.String()
+
+			result, err := k8sBase.Merge(k8sTarget)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).To(BeEquivalentTo(expected))
+		})
+
+		When("configs are empty/invalid", func() {
+			It("returns default when map is empty", func() {
+				extensions := make(map[string]interface{})
```
marcinc, line 192: These 3 lines belong to `BeforeEach` as a stage setup for test cases in this context block. Similar comment goes to the the other case below which would be best to be encapsulated in a separate context.

pkg/kev/config/extensions_test.go
```diff
@@ -124,4 +171,43 @@ var _ = Describe("Extentions", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+
+				result, err := config.DefaultK8SConfig().Merge(parsedConf)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(result).To(BeEquivalentTo(config.DefaultK8SConfig()))
+			})
+
+			It("returns default when map is nil", func() {
+				var extensions map[string]interface{}
+				parsedConf, err := config.K8SCfgFromMap(extensions)
```
marcinc, line 203: it's hard to see from this spec as to what value is meant to be nil? Does the `config.K8SCfgFromMap` ever returns nil? Having a separate context with the stage prep in `BeforeEach` where input nil value is defined would be great for legibility.

marcinc (approved):
LGTM 👍

marcinc (approved):
👍

Issue #458, opened 2021-04-13
rendering manifests from extension values [probes]


mangas:
Liveness and Readiness probes should be rendered based on the presence of ProjectService extensions instead of labels. This will not include parsing from docker-compose into extensions just yet

PR #459, opened 2021-04-14
kev-458 Render probe info from extensions

mangas:
Part 1 of x-k8s extensions - probe rendering

Closes https://github.com/appvia/kev/issues/458


pkg/kev/converter/kubernetes/transform.go
```diff
@@ -1660,20 +1660,16 @@ func (k *Kubernetes) updateKubernetesObjects(projectService ProjectService, obje
 	}
 
 	// @step add PVCs to objects
-	if pvcs != nil {
```
ezodude, line 1663: What changed to allow removing this check?

pkg/kev/converter/kubernetes/transform.go
```diff
@@ -1660,20 +1660,16 @@ func (k *Kubernetes) updateKubernetesObjects(projectService ProjectService, obje
...
-		}
+	// Looping on the slice pvcs instead of `*objects = append(*objects, pvcs...)`
+	// because the type of objects and pvcs is different, but when doing append
+	// one element at a time it gets converted to runtime.Object for objects slice
+	for _, p := range pvcs {
+		*objects = append(*objects, p)
 	}
 
 	// @step add ConfigMaps to objects
-	if cms != nil {
```
ezodude, line 1673: What changed to allow removing this check?

ezodude (commented):
Yay. Love the simplicity and how all the value conversion / checks have been removed.

Left a couple of comments.

pkg/kev/converter/kubernetes/transform.go
mangas: nil is the zero value for slices so for range works on them it just doesn't iterate. No need to protect with a nil check

mangas: https://play.golang.org/p/dB_0-Xoq6HY

pkg/kev/converter/kubernetes/transform.go
mangas: same as the one before

ezodude (approved):
Approving!

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -1254,712 +1240,659 @@ var _ = Describe("ProjectService", func() {
...
-					Expect(err.Error()).To(Equal(fmt.Sprintf("%s needs to be a number", config.LabelWorkloadLivenessProbeTCPPort)))
-					Expect(p).To(BeNil())
-				})
-			})
-		})
-
-	})
-
-	Describe("livenessProbeCommand", func() {
-		When("defined via labels", func() {
```
marcinc, line 1390: There is a bunch of probe related tests being removed in this PR that set the specific contexts around various probe parameters being supplied via configuration (previously labels and now by extensions) and cases for when custom configuration isn't explicitly specified. Do we have the coverage of those things still?

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -1254,712 +1240,659 @@ var _ = Describe("ProjectService", func() {
...
-					_, err := projectService.readinessProbe()
-					Expect(err).NotTo(Succeed())
-					Expect(err.Error()).To(ContainSubstring("%s needs to be a number", config.LabelWorkloadReadinessProbeTCPPort))
+					p, err := projectService.LivenessProbe()
+					Expect(err).NotTo(HaveOccurred())
+					Expect(p.TCPSocket.Port.String()).To(Equal("0"))
 				})
 			})
 
 			Context("and no port label", func() {
```
marcinc, line 1305: It's not a label anymore so let's change the wording here too.

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -1254,712 +1240,659 @@ var _ = Describe("ProjectService", func() {
...
-
-		When("not defined via labels at all", func() {
-			It("returns default value - disable by default", func() {
-				p, err := projectService.readinessProbe()
-				Expect(err).To(Succeed())
-				Expect(p).To(BeNil())
-			})
-		})
-	})
+	// 	Describe("livenessProbeCommand", func() {
```
marcinc, line 1320: Are these commented out specs something we want to keep?

marcinc (commented):
A few additional comments..

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -1159,46 +1172,20 @@ var _ = Describe("ProjectService", func() {
...
 					Handler: v1.Handler{
 						Exec: &v1.ExecAction{
-							Command: healthcheck.Test[1:],
+							Command: config.DefaultLivenessProbeCommand,
 						},
 					},
 					TimeoutSeconds:      10,
-					PeriodSeconds:       10,
-					InitialDelaySeconds: 10,
+					PeriodSeconds:       60,
```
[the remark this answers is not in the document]
mangas: because they are not being read from the compose side yet, so these are the default for now, they will break once compose is read into extensions rather than using defaults

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -1217,34 +1204,33 @@ var _ = Describe("ProjectService", func() {
 				})
 
 				It("logs and returns error", func() {
-					_, err := projectService.healthcheck()
-					Expect(err).To(HaveOccurred())
-					Expect(err).To(MatchError("Health check misconfigured"))
+					_, err := projectService.LivenessProbe()
+					Expect(err).NotTo(HaveOccurred())
+					// Expect(err).To(MatchError("Health check misconfigured"))
```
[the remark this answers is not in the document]
mangas: This needs to be brought back when https://github.com/appvia/kev/issues/460 is completed. Right now the validation is not completed so this doesn't fail but it should.

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -1217,34 +1204,33 @@ var _ = Describe("ProjectService", func() {
...
+					_, err := projectService.LivenessProbe()
+					Expect(err).NotTo(HaveOccurred())
+					// Expect(err).To(HaveOccurred())
+					// Expect(err).To(MatchError("Health check misconfigured"))
 
-					assertLog(logrus.ErrorLevel,
-						"Health check misconfigured",
-						map[string]string{},
-					)
+					// assertLog(logrus.ErrorLevel,
```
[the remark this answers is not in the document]
mangas: yes, same as previous comments

pkg/kev/converter/kubernetes/project_service_test.go
mangas: Yes, we now start with defaults and merge in the set configuration. There are some bits missing in terms of validation (https://github.com/appvia/kev/issues/460) for example 0 values and such but absent values are no longer a problem (and is tested in the config package), same as wrong types

pkg/kev/converter/kubernetes/project_service_test.go
mangas: I'll review all of this once we have values being read and labels completely removed

pkg/kev/converter/kubernetes/project_service_test.go
mangas: yes, we still want to implement them later again but right now there is no point changing them just to change it back later IMO

PR #462, opened 2021-04-16
Provide hooks into the new cmd Runner steps. 

### LINKED ISSUE #461: Provide hooks into the new cmd Runner steps.
This enables other code to extend the runners by listening to pre/post hooks and significant events.

ezodude:
Resolves #461 

App runners now provide hooks to runner steps using runner pre/post and significant events.

See kev/pkg/kev/events.go for a full list of supported events.

**NOTE: this PR is does not include the latest extensions migration work.**


pkg/kev/dev.go
```diff
@@ -33,17 +33,28 @@ import (
...
+	if runner.config.Skaffold && len(runner.config.K8sNamespace) == 0 {
+		runner.config.K8sNamespace = DefaultSkaffoldNamespace
 	}
 	return runner
 }
 
 // Run runs the dev command business logic
 func (r *DevRunner) Run() error {
+	if err := r.eventHandler(DevLoopStarting, r); err != nil {
+		return errors.Errorf("%s\nwhen handling fired event: %s", err.Error(), DevLoopStarting)
```
marcinc, line 55: would it be better if we said which event that was in the message? So `when handling fired event` would become `dev loop couldn't start` or something along these lines?

pkg/kev/dev.go
```diff
@@ -121,8 +137,8 @@ func (r *DevRunner) Run() error {
 				kmd.WithStyle(kmd.LogStyle),
 			)
 
-			if r.chgEventHandler != nil {
-				r.chgEventHandler(ch)
+			if err := r.eventHandler(DevLoopIterated, r); err != nil {
+				return errors.Errorf("%s\nwhen handling fired event: %s", err.Error(), DevLoopIterated)
```
marcinc, line 141: Similar message here. See above.

pkg/kev/init.go
```diff
@@ -54,44 +65,56 @@ func (r *InitRunner) Run() (WritableResults, error) {
...
 	}
 
-	return createInitWritableResults(r.workingDir, r.manifest, skManifest), nil
+	return createInitWritableResults(r.WorkingDir, r.manifest, skManifest), nil
 }
 
 // EnsureFirstInit ensures the project has not been already initialised
 func (r *InitRunner) EnsureFirstInit() error {
+	if err := r.eventHandler(PreEnsureFirstInit, r); err != nil {
+		return errors.Errorf("%s\nwhen handling fired event: %s", err.Error(), PreEnsureFirstInit)
```
marcinc, line 82: And here as well, unless there is a reason why we shouldn't be more explicit? This comment applies to other places where we return error like this.

marcinc (commented):
A couple of minor comments

marcinc (approved):
👍

Issue #443, opened 2021-04-01
Migrate K8s configuration from labels to extenstion


marcinc:
This is an EPIC covering all the work required to transition from label based configuration to extensions.

PR #465, opened 2021-04-20
Revert extension bits

marcinc:
Work related to #443 can render master unstable. This PR reverts all the config extensions work which is now moved to its dedicated feature branch.


ezodude (approved):
👍

PR #467, opened 2021-04-20
kev-458 render extensions

mangas:
This PR adds parsing composego into extensions directly without using some labels. More labels and properties to be added in a following PR

### DIFF
--- a/pkg/kev/config/extensions.go
+++ b/pkg/kev/config/extensions.go
@@ -4,7 +4,10 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"strings"
+	"time"
 
+	composego "github.com/compose-spec/compose-go/types"
 	"github.com/go-playground/validator/v10"
 	"github.com/imdario/mergo"
 	"gopkg.in/yaml.v3"
@@ -75,15 +78,25 @@ func DefaultK8SConfig() K8SConfiguration {
 			ReadinessProbe: DefaultReadinessProbe(),
 			Replicas:       1,
 		},
+		Service: Service{
+			Type: "None",
+		},
 	}
 }
 
 type k8sConfigOptions struct {
 	requireExtensions bool
+	disableValidation bool
 }
 
 // K8SCfgOption will modify parsing behaviour of the x-k8s extension.
 type K8SCfgOption func(*k8sConfigOptions)
+
+func DisableValidation() K8SCfgOption {
+	return func(kco *k8sConfigOptions) {
+		kco.disableValidation = true
+	}
+}
 
 // RequireExtensions will ensure that x-k8s is present and that it is validated.
 func RequireExtensions() K8SCfgOption {
@@ -92,17 +105,122 @@ func RequireExtensions() K8SCfgOption {
 	}
 }
 
-// K8SCfgFromMap handles the extraction of the k8s-specific extension values from the top level map.
-func K8SCfgFromMap(m map[string]interface{}, opts ...K8SCfgOption) (K8SConfiguration, error) {
+func K8SCfgFromCompose(svc *composego.ServiceConfig) (K8SConfiguration, error) {
+	var cfg K8SConfiguration
+
+	cfg.Enabled = true
+	cfg.Workload.Type = WorkloadTypeFromCompose(svc)
+	cfg.Workload.Replicas = WorkloadReplicaFromCompose(svc)
+	cfg.Workload.RestartPolicy = WorkloadRestartPolicyFromCompose(svc)
+	cfg.Service.Type = NoService
+	if svc.Ports != nil {
+		cfg.Service.Type = ClusterIPService
+	}
+
+	cfg.Workload.LivenessProbe = LivenessProbeFromHealthcheck(svc.HealthCheck)
+	cfg.Workload.ReadinessProbe = DefaultReadinessProbe()
+
+	k8sext, err := ParseK8SCfgFromMap(svc.Extensions, DisableValidation())
+	if err != nil {
+		return K8SConfiguration{}, err
+	}
+
+	cfg, err = cfg.Merge(k8sext)
+	if err != nil {
+		return K8SConfiguration{}, err
+	}
+
+	if err := cfg.Validate(); err != nil {
+		return K8SConfiguration{}, err
+	}
+
+	return cfg, nil
+}
+
+func WorkloadRestartPolicyFromCompose(svc *composego.ServiceConfig) string {
+	if svc.Deploy == nil || svc.Deploy.RestartPolicy == nil {
+		return ""
+	}
+
+	switch svc.Deploy.RestartPolicy.Condition {
+	case "on-failure":
+		return RestartPolicyOnFailure
+	case "none":
+		return RestartPolicyNever
+	default:
+		return RestartPolicyAlways
+	}
+}
+
+func WorkloadReplicaFromCompose(svc *composego.ServiceConfig) int {
+	if svc.Deploy == nil || svc.Deploy.Replicas == nil {
+		return 1
+	}
+
+	return int(*svc.Deploy.Replicas)
+}
+
+// TODO: Turn these strings into enums
+func WorkloadTypeFromCompose(svc *composego.ServiceConfig) string {
+	if svc.Deploy != nil && svc.Deploy.Mode == "global" {
+		return DaemonsetWorkload
+	}
+
+	if len(svc.Volumes) != 0 {
+		return StatefulsetWorkload
+	}
+
+	return DeploymentWorkload
+}
+
+func LivenessProbeFromHealthcheck(healthcheck *composego.HealthCheckConfig) LivenessProbe {
+	var res LivenessProbe
+
+	if healthcheck == nil {
+		return DefaultLivenessProbe()
+	}
+
+	if healthcheck.Disable {
+		res.Type = ProbeTypeNone.String()
+		return res
+	}
+
+	res.Type = ProbeTypeExec.String()
+
+	test := healthcheck.Test
+	if len(test) > 0 && strings.ToLower(test[0]) == "cmd" {
+		test = test[1:]
+	}
+	res.Exec.Command = test
+
+	if healthcheck.Timeout != nil {
+		res.Timeout = time.Duration(*healthcheck.Timeout)
+	}
+
+	if healthcheck.Retries != nil {
+		res.FailureThreashold = int(*healthcheck.Retries)
+	}
+
+	if healthcheck.StartPeriod != nil {
+		res.InitialDelay = time.Duration(*healthcheck.StartPeriod)
+	}
+
+	if healthcheck.Interval != nil {
+		res.Period = time.Duration(*healthcheck.Interval)
+	}
+
+	return res
+}
+
+// ParseK8SCfgFromMap handles the extraction of the k8s-specific extension values from the top level map.
+func ParseK8SCfgFromMap(m map[string]interface{}, opts ...K8SCfgOption) (K8SConfiguration, error) {
 	var options k8sConfigOptions
 	for _, o := range opts {
 		o(&options)
 	}
 
 	if _, ok := m[K8SExtensionKey]; !ok && !options.requireExtensions {
-		c := DefaultK8SConfig()
-		c.Workload.Replicas = DefaultReplicaNumber
-		return c, nil
+		return K8SConfiguration{}, nil
 	}
 
 	var extensions ExtensionRoot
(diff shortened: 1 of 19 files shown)


pkg/kev/config/extensions_test.go
```diff
@@ -87,10 +79,14 @@ var _ = Describe("Extentions", func() {
...
-				It("returns an error", func() {
-					parsedCfg, err = config.K8SCfgFromMap(extensions)
-					Expect(err).To(HaveOccurred())
-					Expect(err.Error()).To(ContainSubstring("Workload.Replicas is required"))
+				JustBeforeEach(func() {
+					svc.Extensions = extensions
+				})
+
+				It("returns defaults", func() {
+					parsedCfg, err = config.K8SCfgFromCompose(&svc)
```
marcinc, line 87: A minor comment - the first lines here could be moved up to `JustBeforeEach` block and this `It` could just do the assertion on the `err` and `parsedCfg` values.

pkg/kev/config/extensions_test.go
```diff
@@ -105,38 +101,55 @@ var _ = Describe("Extentions", func() {
...
+					svc.Extensions = extensions
+				})
+
 				When("workload is invalid", func() {
-					It("returns an error", func() {
-						parsedCfg, err = config.K8SCfgFromMap(extensions)
-						Expect(err).To(HaveOccurred())
-						Expect(err.Error()).To(ContainSubstring("Workload.Replicas is required"))
+					It("it is ignored and returns defaults", func() {
+						parsedCfg, err = config.K8SCfgFromCompose(&svc)
```
marcinc, line 110: Similar comment here to the one above. This could be moved to the `JustBeforeEach` block. Also I think it'd be good to set expectation on `parsefCfg` here?

pkg/kev/config/extensions_test.go
```diff
@@ -105,38 +101,55 @@ var _ = Describe("Extentions", func() {
...
+					k8s.Workload.Type = config.DefaultWorkload
 					k8s.Workload.Replicas = 10
 				})
 
 				It("returns error", func() {
-					parsedCfg, err = config.K8SCfgFromMap(extensions)
-					Expect(err).To(HaveOccurred())
-					Expect(err.Error()).To(Equal("Workload.LivenessProbe.Type is required"))
+					parsedCfg, err = config.K8SCfgFromCompose(&svc)
+					Expect(err).NotTo(HaveOccurred())
```
marcinc, line 124: `It` block message says that we expect error but the test here asserts the opposite? Do we also want to check `parsedCfg` is as expected?

pkg/kev/config/extensions_test.go
```diff
@@ -105,38 +101,55 @@ var _ = Describe("Extentions", func() {
...
 					k8s.Workload.LivenessProbe.Type = config.ProbeTypeNone.String()
 				})
 
-				It("returns error", func() {
-					parsedCfg, err = config.K8SCfgFromMap(extensions)
-					Expect(err).To(HaveOccurred())
-					Expect(err.Error()).To(Equal("Workload.Replicas is required"))
+				It("returns in defaults", func() {
+					parsedCfg, err = config.K8SCfgFromCompose(&svc)
+					Expect(err).NotTo(HaveOccurred())
```
marcinc, line 136: I think we're missing the assertion here that verifies the returned value is in fact a default value?

pkg/kev/envs.go
```diff
@@ -197,7 +198,7 @@ func (e *Environment) loadOverride() (*Environment, error) {
 		envVarsFromNilToBlankInService(s)
 		serviceConfig, err := newServiceConfig(s)
 		if err != nil {
-			return nil, errors.Wrapf(err, "Cannot load environment [%s], service [%s]", e.Name, name)
+			log.Debugf("cannot load environment [%s], service [%s]: err %s", e.Name, name, err.Error())
```
marcinc, line 201: I guess we'd still want to return an error here?

pkg/kev/extract.go
```diff
@@ -29,9 +29,9 @@ import (
 
 // extractLabels extracts a set of labels from a compose project object.
 func extractLabels(s composego.ServiceConfig, target *ServiceConfig) {
-	extractServiceTypeLabels(s, target)
-	extractDeploymentLabels(s, target)
-	extractHealthcheckLabels(s, target)
+	// extractServiceTypeLabels(s, target)
```
marcinc, line 32: Should these lines be commented out?

pkg/kev/extract.go
```diff
@@ -75,9 +75,9 @@ func extractServiceTypeLabels(source composego.ServiceConfig, target *ServiceCon
 
 // extractDeploymentLabels extracts deployment related into a label's Service.
 func extractDeploymentLabels(source composego.ServiceConfig, target *ServiceConfig) {
-	extractWorkloadType(source, target)
-	extractWorkloadReplicas(source, target)
-	extractWorkloadRestartPolicy(source, target)
+	// extractWorkloadType(source, target)
```
marcinc, line 78: Should we uncomment those lines?

pkg/kev/init_test.go
```diff
@@ -220,12 +220,16 @@ var _ = Describe("InitRunner", func() {
 		})
 
 		Context("with services", func() {
-			It("should include a subset of labels as config params", func() {
+			It("should include a subset default config params", func() {
```
marcinc, line 223: small rephrase suggestion: `should include default config params`?

pkg/kev/init_test.go
```diff
@@ -220,12 +220,16 @@ var _ = Describe("InitRunner", func() {
...
-				Expect(svc.GetLabels()).To(HaveLen(3))
-				Expect(svc.GetLabels()).To(HaveKey(config.LabelWorkloadLivenessProbeType))
-				Expect(svc.GetLabels()).To(HaveKey(config.LabelWorkloadLivenessProbeCommand))
-				Expect(svc.GetLabels()).To(HaveKey(config.LabelWorkloadReplicas))
+
+				k8sconf, err := config.ParseK8SCfgFromMap(svc.Extensions)
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(svc.GetLabels()).To(BeEmpty())
+				Expect(k8sconf.Workload.LivenessProbe.Type).NotTo(BeEmpty())
```
marcinc, line 230: Should we also assert on value for these things?

pkg/kev/kev_reconcile_test.go
```diff
@@ -261,9 +268,21 @@ var _ = Describe("Reconcile", func() {
 				})
 
 				It("should configure the added service labels from deploy config", func() {
```
marcinc, line 270: The title of this `It` block should also change now?

pkg/kev/kev_reconcile_test.go
```diff
@@ -273,10 +292,17 @@ var _ = Describe("Reconcile", func() {
 				})
 
 				It("should configure the added service labels from healthcheck config", func() {
```
marcinc, line 294: Title of this `It` should change as we don't deal with labels anymore.

pkg/kev/override.go
```diff
@@ -99,9 +99,9 @@ func (o *composeOverride) volumesWithLabelsMatching(other *composeOverride) Volu
 // expandLabelsFrom returns a copy of the compose override
 // filling in gaps in services and volumes labels (keys and values) using the provided override.
 func (o *composeOverride) expandLabelsFrom(other *composeOverride) *composeOverride {
-	services := o.servicesLabelsExpandedFrom(other)
+	// services := o.servicesLabelsExpandedFrom(other)
```
marcinc, line 102: Do we need this line?

pkg/kev/testdata/reconcile-service-deploy/docker-compose.kev.dev.yaml
```diff
@@ -6,8 +6,8 @@ services:
         replicas: 1
         livenessProbe: 
           type: exec
-          exec:
-            command: ["echo", "Define healthcheck command for service db"]
+          # exec:
+          #   command: ["echo", "Define healthcheck command for service db"]
```
marcinc, line 10: Should these be uncommented?

marcinc (commented):
a few comments

pkg/kev/config/extensions.go
```diff
@@ -75,34 +78,149 @@ func DefaultK8SConfig() K8SConfiguration {
...
+	cfg.Enabled = true
+	cfg.Workload.Type = WorkloadTypeFromCompose(svc)
+	cfg.Workload.Replicas = WorkloadReplicaFromCompose(svc)
+	cfg.Workload.RestartPolicy = WorkloadRestartPolicyFromCompose(svc)
+	cfg.Service.Type = NoService
+	if svc.Ports != nil {
+		cfg.Service.Type = ClusterIPService
+	}
+
+	cfg.Workload.LivenessProbe = LivenessProbeFromHealthcheck(svc.HealthCheck)
```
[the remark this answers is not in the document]
mangas: This is not from compose service but from healthcheck

pkg/kev/config/extensions_test.go
```diff
@@ -13,38 +14,38 @@ var _ = Describe("Extentions", func() {
 
 	Describe("parsing", func() {
 		var (
+			k8s config.K8SConfiguration
+			err error
+
+			extensions map[string]interface{}
```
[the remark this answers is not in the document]
mangas: fixed

pkg/kev/config/extensions_test.go
```diff
@@ -87,10 +79,14 @@ var _ = Describe("Extentions", func() {
 					}
 				})
 
-				It("returns an error", func() {
-					parsedCfg, err = config.K8SCfgFromMap(extensions)
-					Expect(err).To(HaveOccurred())
-					Expect(err.Error()).To(ContainSubstring("Workload.Replicas is required"))
+				JustBeforeEach(func() {
+					svc.Extensions = extensions
```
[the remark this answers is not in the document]
mangas: fixed

pkg/kev/config/extensions_test.go
mangas: fixed

pkg/kev/config/extensions_test.go
mangas: fixed

pkg/kev/config/extensions_test.go
mangas: fixed

pkg/kev/envs.go
mangas: no, this one is specifically caused by https://github.com/appvia/kev/blob/efece0d1e0afb7af8adea41fa32d7414d4556e91/pkg/kev/services.go#L32
which is going to be removed at some point

pkg/kev/extract.go
mangas: Yep, I'm progressively removing. these, I'll clean this. up later when they're ready to go

pkg/kev/init_test.go
mangas: updated

pkg/kev/kev_reconcile_test.go
mangas: can we do this later? I'll find all cases and updated as I mentioned earlier.

pkg/kev/kev_reconcile_test.go
mangas: same here

pkg/kev/override.go
mangas: I will remove later on cleanup

pkg/kev/testdata/reconcile-service-deploy/docker-compose.kev.dev.yaml
mangas: yep, well spotted

pkg/kev/config/extensions.go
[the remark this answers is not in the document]
marcinc: OK, but `healthcheck` is part of compose service definition.

pkg/kev/config/extensions_test.go
marcinc: can't fine expectation on `parsedCfg` in this block. I think we should have one as we expect defaults here?

pkg/kev/envs.go
marcinc: OK, but if there is a service config error does the above still make sense? or should we just return immediately with the first error that occurred?

pkg/kev/extract.go
marcinc: OK, maybe just add a TODO comment as it'll be easier to track this later.

pkg/kev/extract.go
marcinc: Let's just leave a TODO comment here if that is temporarily commented out...

pkg/kev/kev_reconcile_test.go
marcinc: OK, I think It's a tiny change and so probably best to address on the spot? As we keep adding those changes it'll be hard to unpick that later I think.

pkg/kev/kev_reconcile_test.go
marcinc: Similar comment to the one above here. Tiny text change so would address on the spot.

pkg/kev/override.go
marcinc: OK, let's just leave a TODO comment here to remember to revisit?

marcinc (commented):
a few more comments

pkg/kev/config/extensions_test.go
marcinc: it'd be good to validate whether `parsedCfg` is as expected though.

pkg/kev/config/extensions.go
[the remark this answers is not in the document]
mangas: maybe the misconception is with the other name, it should probably be FromService or ComposeSvc rather than just "compose". I think establishing a pattern of FromCompose when referring to svc and then FromCompose that takes an healthcheck is confusing and wrong

pkg/kev/envs.go
mangas: for now we move on because we don't care about labels anymore, if it's wrong is because it is either not a label we have ported and in which case it doesn't matter or is one we ported in which case we don't care either so we move on :)

pkg/kev/extract.go
mangas: the function is literally called extractLabels, it will be caught 🤣

pkg/kev/config/extensions.go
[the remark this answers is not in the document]
marcinc: How about `LivenessProbeFromCompose(svc)` then fetching what you need in that function internally?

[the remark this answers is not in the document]
mangas: changed

marcinc (approved):

PR #479, opened 2021-04-22
For UI, use app name defined in the CLI's root cmd

### LINKED ISSUE #475: For UI, use app name defined in the CLI's root cmd
Currently, the app name displayed in the UI depends on the project's manifest.yaml name.

This should not be the case, the app name should either,
- Mirror the name of the app defined in the CLI's root cmd.

ezodude:
Resolves #475 

- The app name is now configured in `main.go`.
- The root cmd is wired to use the app name.
- All runner are now configured to use an app name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,13 @@ else
 	VERSION ?= $(GIT_LAST_TAG)
 endif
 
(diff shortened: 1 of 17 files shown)


marcinc (approved):
LGTM

PR #482, opened 2021-04-23
kev-458 remove label validation

mangas:


pkg/kev/converter/kubernetes/helpers_test.go
```diff
@@ -37,7 +38,7 @@ func init() {
 
 func assertLog(level logrus.Level, message string, fields map[string]string) {
 	Expect(hook.LastEntry().Level).To(Equal(level))
-	Expect(hook.LastEntry().Message).To(Equal(message))
+	Expect(cmp.Diff(hook.LastEntry().Message, message)).To(BeEmpty())
```
ezodude, line 41: Curious on why you had to diff here?

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -32,42 +32,26 @@ import (
...
-
-			return config.DefaultReplicaNumber
-		}
-		return int32(replicas)
-	} else if p.Deploy != nil && p.Deploy.Replicas != nil {
-		return int32(*p.Deploy.Replicas)
-	}
-
-	return config.DefaultReplicaNumber
+	return int32(p.K8SConfig.Workload.Replicas)
```
ezodude, line 54: Yay! 🥳

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -97,39 +98,22 @@ var _ = Describe("ProjectService", func() {
 	})
 
 	Describe("enabled", func() {
-
 		When("component toggle label is set to Truthy value", func() {
```
ezodude, line 101: I think this may have to be reworded as we were previously working with string conversion.

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -291,20 +269,29 @@ var _ = Describe("ProjectService", func() {
...
 
 		Context("when deploy block `mode` defined as `global` and workload type is different than DaemonSet", func() {
-
 			projectWorkloadType := config.StatefulsetWorkload
 
-			BeforeEach(func() {
-				labels = composego.Labels{
-					config.LabelWorkloadType: projectWorkloadType,
-				}
+			JustBeforeEach(func() {
```
ezodude, line 274: No sure why we started using `JustBeforeEach`?

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -1022,10 +1016,19 @@ var _ = Describe("ProjectService", func() {
 			Context("when invalid policy has been provided", func() {
 				policy := "invalid-policy"
 
-				BeforeEach(func() {
-					labels = composego.Labels{
-						config.LabelWorkloadRestartPolicy: policy,
+				JustBeforeEach(func() {
```
ezodude, line 1019: No sure why we started using `JustBeforeEach`?

pkg/kev/services.go
```diff
@@ -18,18 +18,15 @@ package kev
...
 
 	"github.com/appvia/kev/pkg/kev/config"
 	composego "github.com/compose-spec/compose-go/types"
-	"github.com/xeipuuv/gojsonschema"
 )
 
 func newServiceConfig(s composego.ServiceConfig) (ServiceConfig, error) {
 	config := ServiceConfig{Name: s.Name, Labels: s.Labels, Environment: s.Environment, Extensions: s.Extensions}
-	return config, config.validate()
+	return config, nil
```
ezodude, line 29: Do we still need to return an error?

ezodude (commented):
Left some questions/comments.

All in all a great step forward.

pkg/kev/converter/kubernetes/project_service_test.go
marcinc: I guess that's still OK, as the value you can pass in the struct can be anything so good to ensure it's truthy. But since we switched from enabled to the reverse of that some text will need to change for sure.

pkg/kev/config/extensions.go
```diff
@@ -138,18 +138,17 @@ func K8SCfgFromCompose(svc *composego.ServiceConfig) (K8SConfiguration, error) {
...
 
-	switch svc.Deploy.RestartPolicy.Condition {
-	case "on-failure":
-		return RestartPolicyOnFailure
-	case "none":
-		return RestartPolicyNever
-	default:
-		return RestartPolicyAlways
+	if svc.Deploy != nil && svc.Deploy.RestartPolicy != nil {
+		policy = svc.Deploy.RestartPolicy.Condition
```
marcinc, line 148: We shouldn't be returning a literal value from compose here. Previously we were mapping that value to the expected restart policy but this code was removed above.. As it stands this func will be returning mixed values.

pkg/kev/converter/kubernetes/project_service.go
marcinc: That's great, but should probably guard against values passed as strings too? As it stands it'll fail to extrac replicas from the following config:
```yaml
x-k8s:
  workload:
    replicas: "10"
```
Would be good to either support that also, OR handle the error if value is not as expected.

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -146,7 +126,7 @@ func (p *ProjectService) workloadType() string {
 			workloadType)
 	}
 
-	return workloadType
```
marcinc, line 149: I guess no need to change the return value here?

pkg/kev/config/extensions.go
marcinc: Unless this func is meant to return the literal value extracted from compose (if present)? In which case it'd be the extracted value or an empty string, so no priming policy with `RestartPolicyAlways` needed. Does the switch statement need to be moved somewhere else?

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -97,39 +98,22 @@ var _ = Describe("ProjectService", func() {
...
-				}
-			})
-
-			It("returns false", func() {
 				Expect(projectService.enabled()).To(BeFalse())
 			})
 		})
 
-		When("component toggle label is set to any string value not representing a boolean value (one of: 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False.)", func() {
+		When("component toggle extension to set false value", func() {
```
marcinc, line 111: The configuration map may contain values that aren't bool.. eg. "true" , 1, TRUE. It'd make sense to respect and handle those "truthy" values I think.

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -187,11 +166,12 @@ var _ = Describe("ProjectService", func() {
 			})
 
 			It("will use a replica number as specified in deploy block", func() {
+				Expect(projectService.Deploy.Replicas).NotTo(BeNil())
```
marcinc, line 169: Maybe best to assert that the number of replicas is as expected here?

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -1006,6 +999,7 @@ var _ = Describe("ProjectService", func() {
 				})
 
 				It("warns the user and defaults restart policy to `Always`", func() {
+					Expect(projectService.K8SConfig.Workload.RestartPolicy).To(Equal(policy))
```
marcinc, line 1002: Are we expecting workload RestartPolicy defined in x-k8s as "unless-stopped" here i.e. compose notation? or one of the values we defined for it in label based configuration? See: https://github.com/appvia/kev/blob/master/docs/reference/config-params.md#kevworkloadrestart-policy

pkg/kev/kev_reconcile_test.go
```diff
@@ -528,12 +532,13 @@ var _ = Describe("Reconcile", func() {
...
 func newDefaultServiceExtensions(name string, k8sconfs ...config.K8SConfiguration) (map[string]interface{}, error) {
 	k8s := config.K8SConfiguration{
-		Enabled: true,
+		Disabled: false,
 		Workload: config.Workload{
 			LivenessProbe:  config.DefaultLivenessProbe(),
 			ReadinessProbe: config.DefaultReadinessProbe(),
 			Type:           config.DefaultWorkload,
 			Replicas:       config.DefaultReplicaNumber,
+			RestartPolicy:  "always",
```
marcinc, line 541: This should be [config.DefaultRestartPolicy](https://github.com/appvia/kev/blob/master/pkg/kev/config/defaults.go#L29-L30) I reckon.

marcinc (commented):
A few comments

pkg/kev/converter/kubernetes/helpers_test.go
mangas: When I was investigating I couldn't see the whole sentence, just the small difference and it is not useful. This will basically show the 2 sentences in full so it';s easier to understand what changed

pkg/kev/converter/kubernetes/project_service.go
mangas: This will not pass validation because it won't be the right type. At this point we're operating on top of K8SConfiguration which is a struct with defined types. So passing replicas: "10" would result in either a parsing error or a validation error. In either case it is detected before it gets to project_service

pkg/kev/converter/kubernetes/project_service_test.go
mangas: updated

pkg/kev/converter/kubernetes/project_service_test.go
mangas: because we need to change the extensions and in this case the tests needs to override the creation of the project service since it can't be updated easily

pkg/kev/converter/kubernetes/project_service_test.go
mangas: because we need to change the extensions and in this case the tests needs to override the creation of the project service since it can't be updated easily

pkg/kev/services.go
mangas: I would prefer keeping it just because updating the api requires more changes. I think this kind of cleanup should be done afterwards to minimize conflicts with master

pkg/kev/config/extensions.go
mangas: This is the bit where we're moving validation to a different place. The default value used to be handled by project_service but now that doesn't make much sense since this configuration should be responsible for that part. The code as it preserves the functionality since the tests are still covering the same behaviour

pkg/kev/converter/kubernetes/project_service.go
mangas: I reverted this. to use workloadType, this was left because of copy paste. I brought back the logging but forgot to change this last return

pkg/kev/converter/kubernetes/project_service_test.go
mangas: again, we no longer look at maps, we only look at the configuration struct which is type safe so if those values are set wrong they will be either caught by marshalling or validation, this is not the place to check for any of it.

pkg/kev/converter/kubernetes/project_service_test.go
mangas: This check is just to ensure that the value set in `BeforeEach` is being picked up, the following assertion verifies the value so I don't think we need to make this more specific IMO, but if you want I can change it

pkg/kev/converter/kubernetes/project_service_test.go
ezodude: I see.

Personally, to keep spec files simple, I try to only use just one `JustBeforeEach` for the entire spec file. This keeps things readable and easier to understand.

Do you think multiple `JustBeforeEach` statements are code small signalling that we need to split this test file into smaller test files?

pkg/kev/converter/kubernetes/project_service_test.go
ezodude: See https://github.com/appvia/kev/pull/482#discussion_r620190730

pkg/kev/services.go
ezodude: cool.

pkg/kev/converter/kubernetes/project_service_test.go
mangas: k8s should hold whatever is inferred from compose and then merged with extensions content. So both cases can happen.

pkg/kev/kev_reconcile_test.go
mangas: config.DefaultRestartPolicy is "Always" and the specific code from project_service was setting "always" so i left it as it was to avoid adding to the changes. It's definitely something that can be improved later

pkg/kev/converter/kubernetes/project_service_test.go
marcinc: OK, I guess we just need to ensure that what we derive from docker compose is translated into values we want the end user to see/tweak. Currently we let the user to adjust configuration based on the config params we defined - see link above. 

And these are the only values we should be populating in the `projectService.K8SConfig.Workload.RestartPolicy`.  

The spec above expects that the value of `projectService.K8SConfig.Workload.RestartPolicy` is `unless-stopped` which I don't think is correct as it should be one of allowed values: `Always`, `OnFailure`, `Never`.

mangas: so if I understand correctly, you're proposing we move [this](https://github.com/appvia/kev/blob/8621035902e12a1d72078badb617cc5a784d3279/pkg/kev/converter/kubernetes/project_service.go#L457) to the config.K8SCfgFromCompose?

pkg/kev/config/extensions.go
```diff
@@ -148,6 +149,14 @@ func WorkloadRestartPolicyFromCompose(svc *composego.ServiceConfig) string {
 		policy = svc.Deploy.RestartPolicy.Condition
 	}
 
+	if policy == "unless-stopped" {
+		log.WarnWithFields(log.Fields{
+			"restart-policy": policy,
+		}, "Restart policy 'unless-stopped' is not supported, converting it to 'always'")
+
+		policy = "always"
```
marcinc, line 157: OK, this function now returns what has been defined in the docker compose around restart policy, however, we prime the value of policy in the first statement to be `config.RestartPolicyAlways` which is a Kev config parameter value. Perhaps change it to literal string "always" instead?

pkg/kev/converter/kubernetes/project_service_test.go
mangas: Yep, I think so for sure, don't think this is the right time to do it though because of the big refactor in progress but I will definitely add it to the list

ezodude (commented):
👍 Looks good - I'll defer to @marcinc for the final approval (he understands this bit of the code more than I do 😉).

marcinc (approved):
LG. We can address outstanding elements in the follow up PRs.

PR #502, opened 2021-05-04
Extract volume settings into a volumes K8s extension

### LINKED ISSUE #486: Extract volume settings into a volumes K8s extension
- Extract initial settings from Compose Config
-  Consolidate with inferred values from the `converter/kubernetes` pkg - project_services.go#volumes

ezodude:
Resolves #486 and #489

- Introduces volume extensions.

- Ensures volume data is extracted into extensions alongside legacy volume labels.

- Adds extensions volume config validations (resolving #489).

- Uses volume extensions during compose to k8s manifest conversion.

- Ensure `kev init` produces overrides with volume extensions. 
If the compose volume config is minimal, then only defaults will be setup in the produced yaml.


pkg/kev/config/svcextensions_test.go
```diff
@@ -127,7 +143,7 @@ var _ = Describe("Extentions", func() {
 				})
 
 				It("returns in defaults", func() {
-					k8sconf := config.DefaultK8SConfig()
+					k8sconf := config.DefaultSvcK8sConfig()
 					k8sconf.Workload.LivenessProbe.Type = config.ProbeTypeNone.String()
```
marcinc, line 147: Not strictly part of this PR but I think this case could be reworked slightly. Context specifies that we are dealing with missing replicas - but it's hard to see that context being prepared in BeforeEach. Also liveness probe is redundant information here?

pkg/kev/config/volextensions.go
```diff
@@ -0,0 +1,149 @@
...
+	if err := validate.Struct(vkc); err != nil {
+		validationErrors := err.(validator.ValidationErrors)
+		for _, e := range validationErrors {
+			if e.Tag() == "required" {
+				return fmt.Errorf("%s is required", e.StructNamespace())
+			}
+
+			if e.Tag() == "quantity" {
+				return fmt.Errorf(
+					"%s is invalid, use a resource quantity format, e.g. 129M, 10Gi, 123Mi",
```
marcinc, line 83: Maybe simpler examples would be better i.e. `10M, 10Gi, 10Mi`?

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -33,25 +33,25 @@ import (
...
-	cfg, err := config.K8SCfgFromCompose(&svc)
+	cfg, err := config.SvcK8sConfigFromCompose(&svc)
 	if err != nil {
 		return ProjectService{}, err
 	}
 
 	return ProjectService{
 		ServiceConfig: svc,
-		K8SConfig:     cfg,
+		K8sSvc:        cfg,
```
marcinc, line 43: Should this be also changed from `K8sSvc` to `SvcK8sConfig`?

pkg/kev/config/defaults.go
```diff
@@ -24,7 +24,7 @@ const (
 	DefaultVolumeSize = "100Mi"
 
 	// DefaultVolumeStorageClass default PV storage class
-	DefaultVolumeStorageClass = ""
+	DefaultVolumeStorageClass = "standard"
```
marcinc, line 27: We shouldn't make that assumption. Let's leave the default as an empty string as previously.

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -237,30 +237,17 @@ func (p *ProjectService) volumes(project *composego.Project) ([]Volumes, error)
...
-			temp.PVCSize = config.DefaultVolumeSize
-		}
-
-		// set PVC selector from label if present
-		if len(selector) > 0 {
-			temp.SelectorValue = selector
-		}
-
-		// set PVC storage class from label if present, or default class
-		if len(storageClass) > 0 {
```
marcinc, line 258: with default storage class unspecified (empty string) I guess we'd want to keep this conditional?

pkg/kev/init_test.go
```diff
@@ -223,20 +223,32 @@ var _ = Describe("InitRunner", func() {
 			It("should include default config params", func() {
 				svc, _ := env.GetService("db")
 
-				k8sconf, err := config.ParseK8SCfgFromMap(svc.Extensions)
+				k8sSvc, err := config.ParseSvcK8sConfigFromMap(svc.Extensions)
```
marcinc, line 226: would be good to stay consistent and use `svcK8s` instead?

pkg/kev/init_test.go
```diff
@@ -223,20 +223,32 @@ var _ = Describe("InitRunner", func() {
...
 				Expect(svc.GetLabels()).To(BeEmpty())
-				Expect(k8sconf.Workload.LivenessProbe).To(Equal(config.DefaultLivenessProbe()))
-				Expect(k8sconf.Workload.Replicas).NotTo(BeZero())
+				Expect(k8sSvc.Workload.LivenessProbe).To(Equal(config.DefaultLivenessProbe()))
+				Expect(k8sSvc.Workload.Replicas).NotTo(BeZero())
 			})
 		})
-		Context("with volumes", func() {
+
+		Context("with volumes legacy", func() {
```
marcinc, line 235: reword suggestion: `with volume legacy` -> `with volume config labels`

marcinc (commented):
a few minor comments.

pkg/kev/config/svcextensions.go
```diff
@@ -211,7 +227,7 @@ func getMemoryQuantity(b int64) string {
 	return fmt.Sprintf("%.1f%ci", float64(b)/float64(div), "KMGTPE"[exp])
 }
 
-func ImagePullFromCompose(svc *composego.ServiceConfig) (ImagePull, error) {
+func ImagePullFromCompose(_ *composego.ServiceConfig) (ImagePull, error) {
```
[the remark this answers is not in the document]
ezodude: Agreed - I left it as a just in case placeholder.

Will rectify.

pkg/kev/config/svcextensions_test.go
ezodude: Will have a look.

pkg/kev/config/volextensions.go
ezodude: No worries, I think I copied this from K8s docs.

pkg/kev/config/defaults.go
ezodude: Sure. If it's empty, then don't actually need the `DefaultVolumeStorageClass` const at all.

pkg/kev/config/svcextensions_test.go
ezodude: Fixed. IMO there are many BeforeEach / JustBeforeEach combinations making things a bit confusing.

Will aim to refactor those as I go along on new PRs.

pkg/kev/converter/kubernetes/project_service.go
ezodude: Not really, the defaults are now set in `config.VolK8sConfigFromCompose(&composeVol)`.

So if the default is needed, as in storage class didn't start with a value, it will be set there.

PR #516, opened 2021-05-12
Move the pod-security labels to extensions.

### LINKED ISSUE #515: Move the pod-security labels to extensions.
Part of #490

ezodude:
Resolves #515 

- Adds PodSecurity to the extension Workload.
- Changes defaults from `string` to `*int64`s.
- Fixes tests and tidy up.

### DIFF
--- a/pkg/kev/config/defaults.go
+++ b/pkg/kev/config/defaults.go
@@ -45,13 +45,13 @@ const (
 	DefaultImagePullSecret = ""
 
 	// DefaultSecurityContextRunAsUser default UID for pod security context
-	DefaultSecurityContextRunAsUser = ""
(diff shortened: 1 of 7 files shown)


pkg/kev/config/defaults.go
```diff
@@ -45,13 +45,13 @@ const (
 	DefaultImagePullSecret = ""
 
 	// DefaultSecurityContextRunAsUser default UID for pod security context
-	DefaultSecurityContextRunAsUser = ""
+	DefaultSecurityContextRunAsUser = 0
```
marcinc, line 48: We only want to set security context details if provided. In this case `0` represent a valid value which means the default security context will be set wrongly.

marcinc (approved):
LG

PR #519, opened 2021-05-14
Move the rolling update max surge and service account name to extensions.

### LINKED ISSUE #517: Move the rolling update max surge and service account name to extensions.
Part of #490

ezodude:
Resolves #517 

- Moves service account name to k8s extension.
- Ensures rolling update max surge is operational during render.
- Outlines future take for rolling update max surge and max unavailable.
- Fixes tests.


pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -149,15 +149,40 @@ func (p *ProjectService) tlsSecretName() string {
...
+	// *****For now, as we can only configure settings for RollingUpdateMaxSurge, this is the only
+	// strategy that is enabled be default.*****
+	//
+	// An extra setting, e.g. strategy, should be defined to qualify which UpdateStrategy to use. As before,
+	// 'strategy' can be inferred from the Deploy block with defaults added as required.
+	//
+	// The 'strategy' setting will definitely be required when MaxUnavailable config is made available to a k8s extension.
+	// E.g. here's a yaml sample of the proposal:
+	// rollingUpdate:
+	//   strategy: surge
```
marcinc, line 165: If we ever introduce the strategy toggle then we should follow k8s strategies list with `RollingUpdate` being a default. We might need to rethink the way the `strategy` part gets expressed in the k8s extension. It could be as in example below (or similar):
```yaml
strategy:  # will cover Pod replacement / Statefulset update strategies perhaps
  type: RollingUpdate
  rollingUpdate: # there is no additional settings for `Replace` strategy currently.
    maxUnavailable: 2
    maxSurge: 2
```
However, the above is applicable to [Deployments](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy) and we also should rethink this in context of [StatefulSet update strategies](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#update-strategies).

marcinc (commented):
a few comments..

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -326,7 +337,7 @@ func (p *ProjectService) restartPolicy() v1.RestartPolicy {
 			"restart-policy":  policy,
 		}, "Restart policy is not supported, defaulting to 'Always'")
 
-		return v1.RestartPolicy(config.RestartPolicyAlways)
+		return config.RestartPolicyAlways
```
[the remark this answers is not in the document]
ezodude: You don't need the cast, as it's done automatically when the value is returned because the function returns a `v1.RestartPolicy`.

pkg/kev/converter/kubernetes/project_service.go
ezodude: I'll get rid of the comments.

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -149,15 +149,40 @@ func (p *ProjectService) tlsSecretName() string {
...
+	//   strategy: surge
+	//	 maxSurge: 1
+	//   maxUnavailable: 1
+	r := v1apps.RollingUpdateDeployment{}
+
+	if p.SvcK8sConfig.Workload.RollingUpdateMaxSurge > 0 {
+		maxSurge := intstr.FromInt(p.SvcK8sConfig.Workload.RollingUpdateMaxSurge)
+		r.MaxSurge = &maxSurge
+
+		maxUnavailable := intstr.FromInt(0)
```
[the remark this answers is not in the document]
ezodude: Done and tested on local K8s instance.

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -194,6 +198,17 @@ func (p *ProjectService) getKubernetesUpdateStrategy() *v1apps.RollingUpdateDepl
 		return &r
 	}
 
+	if cfg.Order == "start-first" {
+		if cfg.Parallelism != nil {
+			maxSurge := intstr.FromInt(cast.ToInt(*cfg.Parallelism))
+			r.MaxSurge = &maxSurge
+		}
+
+		maxUnavailable := intstr.FromInt(0)
```
marcinc: We could default that to K8s default of 25% here too?

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -522,7 +522,8 @@ var _ = Describe("ProjectService", func() {
 					}
 				})
 
-				expectedMaxUnavailable := intstr.FromInt(0)
+				// expectedMaxUnavailable := intstr.FromInt(0)
```
marcinc, line 525: Line not needed? Not sure if we've got a test covering for setting in a form of an absolute number?

marcinc (commented):
Couple of tiny comments. Otherwise good to go.

pkg/kev/converter/kubernetes/project_service.go
ezodude: I think this might be a GitHub refresh issue, the change has already been made: https://github.com/appvia/kev/pull/519/commits/ed74c6cf7911465bb78f2e91f4526c4dffae9098#diff-178cbdee190dbd6e52b52aad3b5f30f05b300950c8d80fbb647df77a204ac55dR207

pkg/kev/converter/kubernetes/project_service_test.go
ezodude: Will clean up.

In regards to defaults, all tests are checking for defaults the percentage based defaults.

However, the tests also verify the `parallelism` value has been used correctly. This ensures we check for values that are absolute numbers.

marcinc (approved):
LGTM

PR #521, opened 2021-05-14
Move the service labels to extensions

### LINKED ISSUE #520: Move the service labels to extensions
Part of #490

ezodude:
Resolves #520, Depends on PR #519 (this PR is rebased from that branch, specifically the top 3 commits). 

- Completes migration of the x-k8s Service attribute, namely migrating `nodeport` and the `expose` attributes.
See, example
```yaml
service:
  type: ClusterIP
  nodeport:
  expose:
    domain: domain.com
     tlsSecret: my-domain-tls
     ingressAnnotations:
       kubernetes.io/ingress.class: external
       cert-manager.io/cluster-issuer: prod-le-dns01
```

- Ingress now uses `ingressAnnotations`.
- Tidied related conversion workflow.
- Updated tests.


pkg/kev/converter/kubernetes/transform_test.go
```diff
@@ -737,21 +737,30 @@ var _ = Describe("Transform", func() {
 
 		When("project service label instructing to expose the k8s service is specified as empty string", func() {
 			BeforeEach(func() {
-				projectService.Labels = composego.Labels{
-					config.LabelServiceExpose: "",
-				}
+				projectService.SvcK8sConfig.Service.Expose.Domain = ""
+				// projectService.Labels = composego.Labels{
```
marcinc, line 741: can drop commented out lines here and below I think?

pkg/kev/converter/kubernetes/transform_test.go
```diff
@@ -793,13 +803,20 @@ var _ = Describe("Transform", func() {
 		})
 
 		When("project service label instructing to expose the k8s service is specified as `domain.name`", func() {
```
marcinc, line 805: tiny: label -> extension. We can always fish those things out later too.

pkg/kev/converter/kubernetes/transform_test.go
```diff
@@ -793,13 +803,20 @@ var _ = Describe("Transform", func() {
...
+			ingressAnnotations := map[string]string{
```
marcinc, line 806: I guess we could omit annotations in this particular block as we primarily test that domain(s) is set correctly?

pkg/kev/converter/kubernetes/transform_test.go
```diff
@@ -842,15 +860,21 @@ var _ = Describe("Transform", func() {
 		})
 
 		When("project service label instructing to expose the k8s service is specified as comma separated list of domain names", func() {
+			ingressAnnotations := map[string]string{
```
marcinc, line 863: I'd skip annotations here too and focus on domain names only.

pkg/kev/converter/kubernetes/utils.go
```diff
@@ -955,16 +951,6 @@ func contains(strs []string, s string) bool {
 	return i < len(strs) && strs[i] == s
 }
 
-// runtimeObjectConvertToTarget converts runtime object into a target
-func runtimeObjectConvertToTarget(o runtime.Object, target interface{}) error {
```
marcinc, line 959: Just checking.. Wasn't this func in use at all?

pkg/kev/converter/kubernetes/utils.go
```diff
@@ -975,11 +961,6 @@ func ToUnstructured(o runtime.Object) (map[string]interface{}, error) {
 	return raw, nil
 }
 
-// FromUnstructured converts unstructured to target object
-func FromUnstructured(unstructured map[string]interface{}, target interface{}) error {
```
marcinc, line 979: Wasn't this func used? It could have been used downstream perhaps?

pkg/kev/converter/kubernetes/transform_test.go
ezodude: Fixed in #525

pkg/kev/converter/kubernetes/transform_test.go
ezodude: Fixed in #525

pkg/kev/converter/kubernetes/transform_test.go
ezodude: We need the ingress annotations to ensure we create a correct `&networkingv1beta1.Ingress` object.

pkg/kev/converter/kubernetes/transform_test.go
ezodude: Same as above. I'm happy to re-work the tests but I like that we're validating the `&networkingv1beta1.Ingress` object while ensuring the domain is correct.

pkg/kev/converter/kubernetes/utils.go
ezodude: Nope. No where.

Unless `utils.go` has an external entity calling it - nothing was using `runtimeObjectConvertToTarget`.

pkg/kev/converter/kubernetes/utils.go
ezodude: Nope. No where.

Unless `utils.go` has an external entity calling it - nothing was using `FromUnstructured `.

ezodude: After more research nothing uses it.

pkg/kev/converter/kubernetes/utils.go
ezodude: After more research nothing uses it.

pkg/kev/converter/kubernetes/transform_test.go
marcinc: Sure, but I since this particular context block focuses on asserting correctness of domain name(s) I thought removing annotations will make it easier to read and follow. In any case ingress annotations are optional and you should be getting correct `Ingress` object even if they aren't specified. I believe there is another context there testing for annotations injection.

pkg/kev/converter/kubernetes/transform_test.go
marcinc: I'd focus the test around the context and have another case for annotations as well.

pkg/kev/converter/kubernetes/transform_test.go
```diff
@@ -802,134 +802,57 @@ var _ = Describe("Transform", func() {
...
 			})
 
-			It("initialises Ingress with a port routing to the project service name and specifies host information", func() {
-				ing := k.initIngress(projectService, port)
+			It("specifies host in the initialised Ingress", func() {
+				ingress := k.initIngress(projectService, port)
+				Expect(ingress.Spec.Rules[0].Host).To(Equal(domain))
+			})
 
-				Expect(ing).To(Equal(&networkingv1beta1.Ingress{
```
marcinc, line 789: I'm not sure if there is another spec covering the entire ingress object like the one removed here but if not I think I'd be good to add it.

pkg/kev/converter/kubernetes/transform_test.go
```diff
@@ -802,134 +802,57 @@ var _ = Describe("Transform", func() {
...
-						Rules: []networkingv1beta1.IngressRule{
-							{
-								Host: domain,
-								IngressRuleValue: networkingv1beta1.IngressRuleValue{
-									HTTP: &networkingv1beta1.HTTPIngressRuleValue{
-										Paths: []networkingv1beta1.HTTPIngressPath{
-											{
-												Path: "/" + path,
-												Backend: networkingv1beta1.IngressBackend{
-													ServiceName: projectService.Name,
```
marcinc, line 809: I guess we'd need additional case for the port/service too, to make it equivalent to the removed spec?

marcinc (approved):
LG

PR #533, opened 2021-05-20
Turn Restart Policy string value sets into enums

### LINKED ISSUE #493: Turn Restart Policy string value sets into enums
There are a few cases like the following: https://github.com/appvia/kev/blob/5a1c05c568427024fb3ef76a9fc0a5d6bdd1e341/pkg/kev/config/extensions.go#L187

Where the return value is a string and the possible values can only be obtained by reading the code and following all the different options. 

This could be transformed into Enums and supported by a few additional functions which ensure that conversion using different cases works (eg None, noNe, etc) and that there is a Valid() function that can quickly assert validity of a specific string value.

This ticket deals with Restart Policy.
1. Restart Policy

ezodude:
Resolves #493 

- Moved Restart Policy to use enums and clarified inference logic.
- Updated the render phase to use the svc K8s config value directly.
- Added tests to check the various Restart Policy scenarios.

### DIFF
--- a/pkg/kev/config/restart.go
+++ b/pkg/kev/config/restart.go
@@ -76,6 +76,6 @@ func inferRestartPolicyFromComposeValue(v string) RestartPolicy {
 	case "unless-stopped":
 		return RestartPolicyAlways
 	default:
-		return ""
+		return RestartPolicyAlways
 	}
 }
--- a/pkg/kev/config/restart.go
+++ b/pkg/kev/config/restart.go
@@ -1,0 +1,81 @@
+/**
+ * Copyright 2021 Appvia Ltd <info@appvia.io>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package config
+
+import (
+	"strings"
+
+	"github.com/go-playground/validator/v10"
+)
+
+type RestartPolicy string
+
+const (
+	// RestartPolicyAlways default value
+	RestartPolicyAlways RestartPolicy = "Always"
+
+	// RestartPolicyOnFailure restart policy
+	RestartPolicyOnFailure RestartPolicy = "OnFailure"
+
+	// RestartPolicyNever restart policy
+	RestartPolicyNever RestartPolicy = "Never"
+)
+
+// String converts a restart policy to a string value
+func (p RestartPolicy) String() string {
+	return string(p)
+}
+
+// restartPolicies are the only restart policy settings
+var restartPolicies = map[RestartPolicy]bool{
+	RestartPolicyAlways:    true,
+	RestartPolicyOnFailure: true,
+	RestartPolicyNever:     true,
+}
+
+// RestartPoliciesFromValue returns a Restart Policy for a given case insensitive value.
+// Returns a blank string and false for unknown values.
+func RestartPoliciesFromValue(s string) (RestartPolicy, bool) {
+	for k, v := range restartPolicies {
+		if strings.ToLower(k.String()) == strings.ToLower(s) {
+			return k, v
+		}
+	}
+	return "", false
+}
+
+// validateRestartPolicy validator to validate a restart policy
+func validateRestartPolicy(fl validator.FieldLevel) bool {
+	_, valid := RestartPoliciesFromValue(fl.Field().String())
+	return valid
+}
+
+// inferRestartPolicyFromComposeValue infers a Restart Policy for a compose value
+func inferRestartPolicyFromComposeValue(v string) RestartPolicy {
+	switch strings.ToLower(v) {
+	case "", "always", "any":
+		return RestartPolicyAlways
+	case "no", "none", "never":
+		return RestartPolicyNever
+	case "on-failure", "onfailure":
+		return RestartPolicyOnFailure
+	case "unless-stopped":
+		return RestartPolicyAlways
+	default:
+		return ""
+	}
+}
--- a/pkg/kev/config/svcext.go
+++ b/pkg/kev/config/svcext.go
@@ -40,6 +40,38 @@ const (
 
 var dnsSubdomainNameRegex = regexp.MustCompile(dnsSubdomainNamePattern)
 
+type RestartPolicy string
+
+const (
+	// RestartPolicyAlways default value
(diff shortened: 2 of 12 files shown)


pkg/kev/config/svcext.go
```diff
@@ -299,26 +304,18 @@ func getServiceType(serviceType string) (string, error) {
...
-
-	if policy == "unless-stopped" {
-		log.WarnWithFields(log.Fields{
-			"restart-policy": policy,
-		}, "Restart policy 'unless-stopped' is not supported, converting it to 'always'")
-
-		policy = "always"
+func WorkloadRestartPolicyFromCompose(svc *composego.ServiceConfig) RestartPolicy {
+	switch {
+	case svc.Restart != "": // docker-compose v2
```
marcinc, line 309: I don't think it's specific to v2, is it? Also, we should cascade here i.e. the value defined in `Deploy` block should override the value at the top level `Restart`.

pkg/kev/config/restart.go
```diff
@@ -0,0 +1,81 @@
...
+	case "", "always", "any":
+		return RestartPolicyAlways
+	case "no", "none", "never":
+		return RestartPolicyNever
+	case "on-failure", "onfailure":
+		return RestartPolicyOnFailure
+	case "unless-stopped":
+		return RestartPolicyAlways
+	default:
+		return ""
```
marcinc, line 79: this should perhaps return a default `DefaultRestartPolicy` rather than empty string.

pkg/kev/config/svcext_test.go
```diff
@@ -85,88 +82,157 @@ var _ = Describe("Service Extension", func() {
...
-				})
-			})
-
-			Context("invalid/empty workload", func() {
-				JustBeforeEach(func() {
-					svc.Extensions = map[string]interface{}{
-						"x-k8s": map[string]interface{}{
-							"workload": map[string]interface{}{
+			Context("from svc to k8s ext", func() {
+				Context("with missing workload", func() {
```
marcinc, line 86: you meant `with missing workload configuration`?

pkg/kev/config/svcext_test.go
```diff
@@ -85,88 +82,157 @@ var _ = Describe("Service Extension", func() {
...
+					It("returns defaults", func() {
+						defaultWorkload := config.DefaultSvcK8sConfig().Workload
+						Expect(parsedK8sCfg.Workload).To(BeEquivalentTo(defaultWorkload))
+					})
 				})
 
-				When("workload is invalid", func() {
-					It("it is ignored and returns defaults", func() {
-						Expect(parsedK8sCfg).To(BeEquivalentTo(config.DefaultSvcK8sConfig()))
+				Context("invalid/empty workload", func() {
```
marcinc, line 101: tiny nitpick, it'd be nice to keep the same wording as in previous context: `with invalid/empty workload configuration`. Should we tweak other context titles below in line with it it?

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -322,22 +322,22 @@ func (p *ProjectService) serviceAccountName() string {
...
-	restartPolicy, err := getRestartPolicy(p.Name, policy)
-	if err != nil {
-		log.WarnWithFields(log.Fields{
-			"project-service": p.Name,
-			"restart-policy":  policy,
-		}, "Restart policy is not supported, defaulting to 'Always'")
-
-		return config.RestartPolicyAlways
-	}
+	// policy := config.RestartPolicyAlways
```
marcinc, line 325: comments..

pkg/kev/testdata/init-default/docker-compose-yml/docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3.9'
+version: '2.4'
```
marcinc, line 1: Why this change?

marcinc (commented):
A few comments. LGTMO. Also, liking the tweaks and cleanup to the test coverage 👍

pkg/kev/config/svcext.go
ezodude: Nope ... Its a v2 vs v3 thing. No cascade required. Hence the re-implementation.

v2: https://docs.docker.com/compose/compose-file/compose-file-v2/#restart
v3: https://docs.docker.com/compose/compose-file/compose-file-v3/#restart_policy

pkg/kev/testdata/init-default/docker-compose-yml/docker-compose.yml
ezodude: `restart: always` is a v2 docker-compose type config not v3.

> deploy.restart_policy  ... Configures if and how to restart containers when they exit. Replaces restart.
See here: https://docs.docker.com/compose/compose-file/compose-file-v3/#restart_policy

pkg/kev/config/svcext.go
marcinc: I asked as `Restart` is also possible in [V3 manifest](https://docs.docker.com/compose/compose-file/compose-file-v3/#restart), and thus we should still cascade I think.

pkg/kev/testdata/init-default/docker-compose-yml/docker-compose.yml
marcinc: see my comment below. `Restart` is still a valid v3 field.

pkg/kev/config/svcext.go
ezodude: That's odd the spec states otherwise.

ezodude: Corrected.

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -320,24 +321,18 @@ func (p *ProjectService) serviceAccountName() string {
...
-			"restart-policy":  policy,
-		}, "Restart policy is not supported, defaulting to 'Always'")
+// restartPolicy returns workload restart policy
+func (p *ProjectService) restartPolicy() (v1.RestartPolicy, error) {
+	return toV1RestartPolicy(p.SvcK8sConfig.Workload.RestartPolicy)
+}
 
-		return config.RestartPolicyAlways
+// toV1RestartPolicy maps to a case-sensitive v1 restart policy
+func toV1RestartPolicy(rp config.RestartPolicy) (v1.RestartPolicy, error) {
```
marcinc, line 330: should this func be moved to `restart.go`? I think this func could just take any previously inferred or config extracted RestartPolicy value and ensure that it returns a valid v1.RestartPolicy. In case the current config value wasn't correct perhaps we could just fall back to default instead of erroring? Though I get why you'd want to error in some cases... What do you think?

marcinc (commented):
one last comment.

pkg/kev/converter/kubernetes/project_service.go
ezodude: Funny you should ask this question - because that's where the func started 😄 ...

But I moved it here because its a converter issue. Different converters may need to deal with this value in different ways.

Regarding the error, the likelihood is small but it's a useful guard. Also, useful for a user to rectify any bad config.

marcinc (approved):
LG

PR #534, opened 2021-05-21
Turn Workload Type and Service Type value sets into enums

### LINKED ISSUE #532: Turn Workload Type and Service Type value sets into enums
There are a few cases like the following: https://github.com/appvia/kev/blob/5a1c05c568427024fb3ef76a9fc0a5d6bdd1e341/pkg/kev/config/extensions.go#L187

Where the return value is a string and the possible values can only be obtained by reading the code and following all the different options. 

This could be transformed into Enums and supported by a few additional functions which ensure that conversion using different cases works (eg None, noNe, etc) and that there is a Valid() function that can quickly assert validity of a specific string value.

This ticket deals with Restart Policy.

- [x] 1. Workload Type

- [x] 2. Service Type 

Should review the whole configuration tree

ezodude:
Resolves #532 

- Workload Types are now enums, and can handle case insensitive k8s extension config.
- Service Types are now enums, and can handle case insensitive k8s extension config.

### DIFF
--- a/pkg/kev/config/svcext.go
+++ b/pkg/kev/config/svcext.go
@@ -84,6 +84,10 @@ func (skc SvcK8sConfig) Validate() error {
 		return err
 	}
 
+	if err := validate.RegisterValidation("workloadType", validateWorkloadType); err != nil {
+		return err
+	}
+
 	if err := validate.RegisterValidation("restartPolicy", validateRestartPolicy); err != nil {
 		return err
 	}
@@ -326,14 +330,13 @@ func WorkloadReplicasFromCompose(svc *composego.ServiceConfig) int {
 	return int(*svc.Deploy.Replicas)
 }
 
-func WorkloadTypeFromCompose(svc *composego.ServiceConfig) string {
-	// TODO: Turn these strings into enums
+func WorkloadTypeFromCompose(svc *composego.ServiceConfig) WorkloadType {
 	if svc.Deploy != nil && svc.Deploy.Mode == "global" {
-		return DaemonsetWorkload
+		return DaemonSetWorkload
 	}
 
 	if len(svc.Volumes) != 0 {
-		return StatefulsetWorkload
+		return StatefulSetWorkload
 	}
 
 	return DeploymentWorkload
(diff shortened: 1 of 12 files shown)


pkg/kev/config/svcext.go
```diff
@@ -260,18 +267,18 @@ func PodSecurityWithDefaults() PodSecurity {
 	}
 }
 
-func ServiceTypeFromCompose(svc *composego.ServiceConfig) (string, error) {
-	serviceType := NoService
+func ServiceTypeFromCompose(svc *composego.ServiceConfig) (ServiceType, error) {
+	var candidate = "none"
```
marcinc, line 271: What's the reason for this change (and the other cases below)?

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -85,11 +85,11 @@ func (p *ProjectService) workloadType() string {
 }
 
 // serviceType returns service type for project service workload
-func (p *ProjectService) serviceType() (string, error) {
+func (p *ProjectService) serviceType() (config.ServiceType, error) {
 	serviceType := p.SvcK8sConfig.Service.Type
 
 	// @step validate whether service type is set properly when node port is specified
-	if !strings.EqualFold(serviceType, string(v1.ServiceTypeNodePort)) && p.nodePort() != 0 {
+	if !config.ServiceTypesEqual(serviceType, config.NodePortService) && p.nodePort() != 0 {
```
marcinc, line 92: minor comment, `v1.ServiceTypeNodePort` is unlikely to change in contrast to the value of `config.NodePortService`... I know they both represent the same thing but perhaps we should stick to the former?

marcinc (commented):
couple of tiny comments.

pkg/kev/config/svcext.go
ezodude: I've reverted to using the raw values determined by the various rules.

Then, `inferServiceTypeFromComposeValue` can use its own internal rules to turn the candidate raw value into the actual `ServiceType` enum.

Otherwise, it can get confusing as to why I'm using an enum to start with. Changing it to a string, and then `inferServiceTypeFromComposeValue` changing it back to an enum.

marcinc (approved):

PR #536, opened 2021-05-25
Minify Extensions

### LINKED ISSUE #497: Minify Extensions
With the change to extensions kev lost the ability to output only a few properties to the overlay file. 

Right now this is not possible because reconcile will only look at the values of composego.ServiceConfig.Extensions

https://github.com/appvia/kev/blob/5a1c05c568427024fb3ef76a9fc0a5d6bdd1e341/pkg/kev/manifest.go#L208

https://github.com/appvia/kev/blob/5a1c05c568427024fb3ef76a9fc0a5d6bdd1e341/pkg/kev/override.go#L188

Changing this behaviour could have some unexpected behaviour like the Inferred value from the composego.ServiceConfig in docker-compose.yaml being different from the previously rendered one in a way that makes the manifest non upgradeable. I am not sure if this is something that will happen in practice but I think it's noting here. 

The goal for this ticket is to find a way to remove the values that are the same when comparing the result of 

K8SCfgFromCompose and ParseK8SCfgFromMap since those could be "inflated" in a following command. 

For this I had thought of changing the code of mergo.Merge to essentially remove the *dst.Property when dst.Property== src.Property, this combined with omitempty tags for whatever properties are not required to show in the overlay should achieve the goal.

ezodude:
This resolves #497 

- Minifies environment overrides on init - including overwrite values received from the sources.

- Applies a just in time configuration strategy by overlaying config info at the last minute to render the required k8s manifests.

Config info is used from:
- Sources compose config.
- Any sources' k8s extensions.
- Minified environment overrides.

### DIFF
--- a/pkg/kev/config/minify.go
+++ b/pkg/kev/config/minify.go
@@ -16,8 +16,8 @@
 
 package config
 
-// MinifySvcExtension creates a minimal service extension configuration using the supplied src.
-func MinifySvcExtension(src map[string]interface{}) (map[string]interface{}, error) {
+// MinifySvcK8sExtension creates a minimal service extension configuration using the supplied src.
+func MinifySvcK8sExtension(src map[string]interface{}) (map[string]interface{}, error) {
 	srcCfg, err := ParseSvcK8sConfigFromMap(src)
 	if err != nil {
 		return nil, err
@@ -50,8 +50,8 @@ func MinifySvcExtension(src map[string]interface{}) (map[string]interface{}, err
 	}.Map()
 }
 
-// MinifyVolExtension creates a minimal volume extension configuration using the supplied src.
-func MinifyVolExtension(src map[string]interface{}) (map[string]interface{}, error) {
+// MinifyVolK8sExtension creates a minimal volume extension configuration using the supplied src.
+func MinifyVolK8sExtension(src map[string]interface{}) (map[string]interface{}, error) {
 	srcCfg, err := ParseVolK8sConfigFromMap(src)
 	if err != nil {
 		return nil, err
--- a/pkg/kev/config/minify.go
+++ b/pkg/kev/config/minify.go
@@ -1,0 +1,59 @@
+/**
+ * Copyright 2021 Appvia Ltd <info@appvia.io>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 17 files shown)


pkg/kev/config/minify.go
```diff
@@ -0,0 +1,61 @@
...
+		probeConfig = ProbeConfig{
+			HTTP: srcCfg.Workload.LivenessProbe.ProbeConfig.HTTP,
+		}
+	case ProbeTypeTCP.String():
+		probeConfig = ProbeConfig{
+			TCP: srcCfg.Workload.LivenessProbe.ProbeConfig.TCP,
+		}
+	}
+
+	return SvcK8sConfig{
```
marcinc, line 42: will JIT config respect any additional parameters user may include in environment override?

pkg/kev/manifest.go
```diff
@@ -181,39 +188,83 @@ func (m *Manifest) ReconcileConfig(envs ...string) (*Manifest, error) {
...
+func validateEnvExtensions(e *Environment, base *composeOverride) error {
+	for _, s := range e.GetServices() {
+		baseSvc, missingSvcErr := base.getService(s.Name)
+		if missingSvcErr != nil {
+			continue
+		}
+
+		baseSvcK8sCfg, err := config.ParseSvcK8sConfigFromMap(baseSvc.Extensions, config.SkipValidation())
+		if err != nil {
+			return errors.Wrapf(missingSvcErr, "when parsing service %s extensions in base override", baseSvc.Name)
```
marcinc, line 227: is it base or override? Am not sure I understand this message correctly.

pkg/kev/manifest.go
```diff
@@ -181,39 +188,83 @@ func (m *Manifest) ReconcileConfig(envs ...string) (*Manifest, error) {
...
+
+	for name, vol := range e.GetVolumes() {
+		baseVol, missingVolError := base.getVolume(name)
+		if missingVolError != nil {
+			continue
+		}
+
+		baseVolK8sCfg, err := config.ParseVolK8sConfigFromMap(baseVol.Extensions)
+		if err != nil {
+			return errors.Wrapf(missingVolError, "when parsing vol %s extensions in base override", name)
```
marcinc, line 253: similar comment here regarding "base override". I'm guessing we are talking about volume k8s config (via extension) in base compose file?

marcinc (commented):
a few small comments

pkg/kev/config/minify.go
ezodude: Yes, they get merged during render.

pkg/kev/manifest.go
ezodude: Yes that's exactly right: volume k8s config (via extension) in base compose file.

pkg/kev/manifest.go
ezodude: Same as volumes below. It's the service k8s config (via extension) in the base compose file.

pkg/kev/manifest.go
marcinc: That's what I figured out. Feels like "override" at the end is superfluous?

marcinc (approved):
👍   I left one tiny comment there but overall LGTM.

marcinc (approved):
👍

PR #538, opened 2021-05-26
Improves skaffold manifest build artifacts detection with support for build strategy with CNB

### LINKED ISSUE #537: Apply correct strategy to Skaffold build artifacts
Kev detects project's artefacts and populates Skaffold manifest with them. Previously, build artefacts were picked up by Skaffold analysis and/or extracted from Docker Compose manifest. This approach assumed that Dockerfile (to create a docker image with) is present and therefore a local docker build strategy was used by default. 

However, there may be situations in which Dockerfile isn't present in the project directory and the docker compose manifest references a docker image along with build context specified. Kev should recognise that use case and generate Skaffold build artefacts configuration that uses Cloud Native Buildpacks as building method for such artefacts.

Another common case could be that Skaffold analysis doesn't detect any Dockerfiles / docker images in the application source code, and docker compose application manifest specifies docker image explicitly, however, without build context specified. This usually means that application utilises pre-built (ready to use) images that don't need to be built off of the project source code. 

(edge case) Final scenario covers the situation where Skaffold analysis doesn't detect any Dockerfile and docker compose manifest doesn't reference any image. In such case assume image name as local project name and build from source with buildpacks (?) NOTE: docker compose requires presence of image and/or build context, otherwise it'll fail configuration check.

marcinc:
Resolves #537 

This PR slightly modifies the way Skaffold build artifacts are detected, and as a result strategy used to build them.
By default a `local docker build` strategy is used, unless no Dockerfile has been detected and docker compose manifest (application stack representation) references docker images with a build context - this suggest that artefact build should still go ahead, however, as there is no Dockerfile present it should use CNB to do the build.

### DIFF
--- a/pkg/kev/skaffold.go
+++ b/pkg/kev/skaffold.go
@@ -81,11 +81,8 @@ var (
 // NewSkaffoldManifest returns a new SkaffoldManifest struct.
 func NewSkaffoldManifest(envs []string, project *ComposeProject) (*SkaffoldManifest, error) {
 
-	analysis, err := analyzeProject()
-	if err != nil {
-		// just warn for now - potentially put project analysis behind a flag?
-		log.Warn(err.Error())
-	}
+	// it's OK to pass nil analysis so no error handling necessary here
+	analysis, _ := analyzeProject()
 
 	manifest := BaseSkaffoldManifest()
 	manifest.SetBuildArtifacts(analysis, project)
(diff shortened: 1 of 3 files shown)


pkg/kev/skaffold.go
```diff
@@ -81,11 +81,8 @@ var (
 // NewSkaffoldManifest returns a new SkaffoldManifest struct.
 func NewSkaffoldManifest(envs []string, project *ComposeProject) (*SkaffoldManifest, error) {
```
ezodude, line 82: `NewSkaffoldManifest` only returns a `nil` error. Maybe remove the returned `error` arg.

pkg/kev/skaffold_test.go
```diff
@@ -521,11 +513,70 @@ var _ = Describe("Skaffold", func() {
...
+								BuildpackArtifact: &latest.BuildpackArtifact{
+									Builder: "paketobuildpacks/builder:base",
+								},
+							}))
+						})
+					})
+
 				})
 
 				When("Docker Compose project doens't have services referencing images with build contexts", func() {
```
ezodude, line 582: `doens't` -> `doesn't` 😉

ezodude (commented):
Looks good so far. Left 2 small comments.

ezodude (approved):
👍

PR #547, opened 2021-06-01
Update the documentation to use extensions instead of labels.

### LINKED ISSUE #546: Update the documentation to use extensions instead of labels.

ezodude:
Resolves #546 

- Updates documentation to use `x-k8s` instead of labels.
- Introduces the concept of project wide configuration.
- Has the latest from `master`.


docs/reference/config-example.md
```diff
@@ -10,59 +10,81 @@ version: "3.7"
...
+          ### TCP
+          type: tcp                                      # TCP readiness probe type.
+          tcp:
+            port: 8080                                   # TCP readiness probe port. Only used when using an tcp probe type.
+          failureThreshold: 3                            # Default: 3. Number of probe retries. Shared by all readiness probe types.
+          initialDelay: 1m0s                             # Default: 1m. How long to wait before initial probe run. Shared by all readiness probe types.
+          period: 1m0s                                   # Default: 1m. How often readiness probe should run for the workload. Shared by all readiness probe types.
+          timeout: 10s                                   # Default: 10s. Probe timeout. Shared by all readiness probe types.
+        replicas: 25                                     # Default: 1. Number of replicas per workload.
+        resource:                                        # Resource share request for a given workload
```
marcinc, line 70: Let's rephrase to `Defines resource requests and limits`? Could also reorder memory / maxMemory.

docs/reference/config-example.md
```diff
@@ -10,59 +10,81 @@ version: "3.7"
...
+          cpu: "0.1"                                     # Default: 0.1. CPU request per workload.
+          maxCpu: "0.5"                                  # Default: 0.5. CPU limit per workload.
+          maxMemory: 500Mi                               # Default: 500Mi. Memory limit per workload.              
+          memory: 10Mi                                   # Default: 10Mi. Memory request per workload.
+        restartPolicy: Always                            # Default: Always. Possible options: Always / OnFailure / Never.
+        rollingUpdateMaxSurge: 1                         # Default: 1. Maximum number of containers to be updated at a time.
+        serviceAccountName: default                      # Default: default. Service account to be used.
```
marcinc, line 77: `Service account name ...` for clarity?

docs/reference/config-params.md
```diff
@@ -5,196 +5,227 @@ title: Kev configuration reference
 
 # Configuration
 
-Kev leverages Docker Compose specification to configure and prepare an application for deployment in Kubernetes. Environment configuration lives in a dedicated docker compose override file, which automatically gets applied to the project's source sdocker compose files at `render` phase.
+Kev leverages the Docker Compose specification to configure and prepare an application for deployment in Kubernetes.
+
+## Project wide configuration
+
+This is optional, but if provided, lives in the base docker compose files (or at least one of the files) used to initialise the project.
```
marcinc, line 12: tiny reword suggestion: `Project wide configuration is OPTIONAL and can be defined in one (or more) of the source compose files used to initialise the project.` ... followed with `It will be applied against ..`

docs/reference/config-params.md
```diff
@@ -5,196 +5,227 @@ title: Kev configuration reference
...
 ### Component level configuration
 
 Configuration is divided into the following groups of parameters:
-* [Component](#-component)
-* [Workload](#-workload)
-* [Service](#-service)
-* [Volumes](#-volumes)
-* [Environment](#-environment)
+
+* [Component](#-component), part of the `x-k8s` docker compose service extension.
```
marcinc, line 26: perhaps instead of duplicating the same comment, we can put a top level item grouping these things together? Not sure if we need those extra comments here as the configuration settings will become apparent in their respective sections.

docs/reference/config-params.md
```diff
@@ -5,196 +5,227 @@ title: Kev configuration reference
...
+* [Volumes](#-volumes),  part of the `x-k8s` docker compose volume extension.
+* [Environment](#-environment),  part of the docker compose environment service attributes.
 
 # → Component
 
-This configuration group contains application composition related settings. Configuration parameters can be individually defined via set of labels (listed below) for each application stack component.
+This configuration group contains application composition related settings. Configuration parameters can be individually defined for each application stack component.
 
-## kev.component.enabled
+## x-k8s.disabled
```
marcinc, line 36: Should we drop `x-k8s.` prefix here and in all cases below?

docs/reference/config-params.md
```diff
@@ -5,196 +5,227 @@ title: Kev configuration reference
...
 
 Defines docker image pull policy from the container registry. See official K8s [documentation](https://kubernetes.io/docs/concepts/containers/images/#updating-images).
 
-### Default: `IfNotPresent`
+#### Default: `IfNotPresent`
 
-### Possible options: `IfNotPresent`, `Always`.
+#### Possible options: `IfNotPresent`, `Always`, `Never`.
 
 > kev.workload.image-pull-policy:
```
marcinc, line 70: header will change now I guess?

docs/reference/config-params.md
```diff
@@ -5,196 +5,227 @@ title: Kev configuration reference
...
-> kev.workload.service-account-name:
+> x-k8s.workload.serviceAccountName:
 ```yaml
 version: 3.7
 services:
   my-service:
-    labels:
-      kev.workload.service-account-name: my-special-service-account-name
+    x-k8s:
+    workload:
```
marcinc, line 135: indent issue

docs/reference/config-params.md
```diff
@@ -5,196 +5,227 @@ title: Kev configuration reference
...
+      podSecurity:
+        runAsGroup: 2000
 ...
 ```
 
-## kev.workload.pod-security-fs-group
+### x-k8s.workload.podSecurity.fsGroup
 
 Defines the [Pod Security Context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for the kubernetes workload. This option is concerned with setting up a supplementary group `fsGroup` field. If specified, all processes of the container are also part of this supplementary group ID. The owner for attached volumes and any files created in those volume will be Group ID as specified by the value of this configuration option.
```
marcinc, line 184: We could maybe drop the opening `Defines the [Pod Security ...]...` sentence as covered in the top block?

docs/reference/config-params.md
```diff
@@ -203,75 +234,88 @@ Otherwise, number of replicas will default to `1`.
...
 ```yaml
 version: 3.7
 services:
   my-service:
-    labels:
-      kev.workload.autoscale-cpu-threshold: 80
+    x-k8s:
+      workload:
+        autoscale:
+          maxReplicas: 10
```
marcinc, line 288: maybe add a comment here that assumes replicas to be set to a value lower than 10? Similar comment in the example below.

docs/reference/config-params.md
```diff
@@ -280,462 +324,543 @@ Otherwise it will default to `1`.
...
+    x-k8s:
+      workload:
+        rollingUpdateMaxSurge: 2
 ...
 ```
 
-## kev.workload.cpu
+## x-k8s.workload.resource
+
+Defines the resource share request for a given workload using different parameters.
```
marcinc, line 340: we should mention both: requests and limits

docs/reference/config-params.md
```diff
@@ -280,462 +324,543 @@ Otherwise it will default to `1`.
...
+        resource:
+          cpu: 0.1
 ...
 ```
 
-## kev.workload.max-cpu
+### x-k8s.workload.resource.maxCpu
 
-Defines the CPU share limit for a given workload. See official K8s [documentation](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/). Kev will attempt to infer CPU request from the information specified in the compose file.
+Defines the max CPU share limit for a given workload. See official K8s [documentation](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/). Kev will attempt to infer CPU request from the information specified in the compose file.
```
marcinc, line 368: I think the `limit` sets the boundaries here so probably can omit added `max` here?

docs/reference/config-params.md
```diff
@@ -280,462 +324,543 @@ Otherwise it will default to `1`.
...
 
-Kev uses the following heuristics to derieve that information for each service:
+Kev uses the following heuristics to derive that information for each service:
 
 If compose file(s) specifies the `deploy.resources.limits.cpus` attribute key in a service config it will use its value.
 Otherwise it'll default to a sensible default of `0.2` (equivalent of 200m in Kubernetes).
 
-### Default: `0.2`
+#### Default: `0.5`
```
marcinc, line 375: Is it `0.5` now? If so we should update the copy above too.

docs/reference/config-params.md
```diff
@@ -280,462 +324,543 @@ Otherwise it will default to `1`.
...
+        resource:
+          memory: 200Mi
 ...
 ```
 
-## kev.workload.max-memory
+### x-k8s.workload.resource.maxMemory
 
-Defines the Memory limit for a given workload. See official K8s [documentation](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/). Kev will attempt to infer Memory limit from the information specified in the compose file.
+Defines the max Memory limit for a given workload. See official K8s [documentation](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/). Kev will attempt to infer Memory limit from the information specified in the compose file.
```
marcinc, line 417: `max` superfluous here as `Memory limit` implies it's a maximum value?

docs/reference/config-params.md
```diff
@@ -280,462 +324,543 @@ Otherwise it will default to `1`.
...
+See official K8s [documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command). Kev will attempt to infer the command from the information specified in the compose file.
 
-Kev uses the following heuristics to derieve that information for each service:
+Kev uses the following heuristics to derive that information for each service:
 
 If compose file(s) specifies the `healthcheck.test` attribute key in a service config it will use its value.
 If probe is not defined it will prompt the user to define one by injecting generic echo command.
 
-### Default: echo "prompt user to define the probe"
+#### Default: echo "prompt user to define the probe"
```
marcinc, line 479: Is it still a default?

docs/reference/config-params.md
```diff
@@ -280,462 +324,543 @@ Otherwise it will default to `1`.
...
+    x-k8s:
+      workload:
+        livenessProbe:
+          type: tcp
+          tcp:
+            port: 8080
 ...
 ```
 
-## kev.workload.liveness-probe-interval
```
marcinc, line 480: Was `interval` section accidentally removed?

docs/reference/config-params.md
```diff
@@ -280,462 +324,543 @@ Otherwise it will default to `1`.
...
+Defines the liveness probe's failure threshold (retries) for the workload. See official K8s [documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command). Kev will attempt to infer the timeout value from the information specified in the compose file.
 
-If compose file(s) specifies the `healthcheck.interval` attribute key in a service config it will use its value.
-Otherwise it'll default to `1m` (1 minute).
-
-### Default: `1m`
+#### Default: `3`
 
-### Possible options: Time duration
+#### Possible options: Arbitrary time duration. Example: `5`
```
marcinc, line 568: I guess this was applicable to `interval` setting? Perhaps sections got mixed up a little?

docs/reference/config-params.md
```diff
@@ -280,462 +324,543 @@ Otherwise it will default to `1`.
...
-If compose file(s) specifies the `healthcheck.start_period` attribute key in a service config it will use its value.
-Otherwise it'll default to `1m` (1 minute).
+If compose file(s) specifies the `healthcheck.interval` attribute key in a service config it will use its value.
+Otherwise, it'll default to `1m` (1 minute).
 
-### Default: `1m`
+#### Default: `1m`
 
-### Possible options: Arbitrary time duration. Example: `1m30s`
+#### Possible options: Time duration
```
marcinc, line 620: Adding example of time duration would be helpful.

docs/reference/config-params.md
```diff
@@ -280,462 +324,543 @@ Otherwise it will default to `1`.
...
+If compose file(s) specifies the `healthcheck.start_period` attribute key in a service config it will use its value.
+Otherwise, it'll default to `1m` (1 minute).
 
-### Default: `3`
+#### Default: `1m`
 
-### Possible options: Arbitrary integer. Example: `5`
+#### Possible options: Arbitrary time duration. Example: `1m30s`
 
-> kev.workload.liveness-probe-retries:
```
marcinc, line 516: Has `retries` been covered with `failureThreshold`? or skipped accidentally?

docs/reference/config-params.md
```diff
@@ -280,462 +324,543 @@ Otherwise it will default to `1`.
...
+    x-k8s:
+      workload:
+        livenessProbe:
+          ...
+          timeout: 30s
 ...
 ```
 
-## kev.workload.readiness-probe-type
+## x-k8s.workload.readinessProbe
```
marcinc, line 661: `readinessProbe` settings will match those of livenessProbe so worth checking whether they are on par. Alternatively could rework that section and point at relevant `livenessProbe` config sections with a note that `readinessProbe` should be used as top level key?

docs/reference/config-params.md
```diff
@@ -805,124 +932,157 @@ Again, it is ideal for exposing a service or app to the internet under a single
...
+## x-k8s.service.expose
 
-Defines whether to expose the service to external world. This detail can't be easily derived from the compose file and so in order to expose a service to external world user must explicitly instruct Kev to do so. By default all component services aren't exposed i.e. have no ingress attached to them.
+Defines how to expose the service externally. This detail can't be easily derived from the compose file and so in order to expose a service the user must explicitly instruct Kev to do so. By default, all component services aren't exposed i.e. have no ingress attached to them.
 
-### Default: `""` - No ingress will be created!
+### x-k8s.service.expose.domain
 
-### Possible options:
-* `"true"` - ingress will be created with Kubernetes cluster defaults
```
marcinc, line 836: I think it's useful to leave that info in the docs.

docs/reference/config-params.md
```diff
@@ -805,124 +932,157 @@ Again, it is ideal for exposing a service or app to the internet under a single
...
-## kev.service.expose.tls-secret
+### x-k8s.service.expose.tlsSecret
 
 Defines whether to use TLS for the exposed service and which secret name contains certificates for the service. See official K8s [documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls).
 
-NOTE: This option is only relevant when service is exposed, see: [kev.service.expose](#kev-service-expose) above.
+NOTE: This option is only relevant when service is exposed, see: [x-k8s.service.expose.domain](#x-k8s.service.expose.domain) above.
 
-### Default: `nil` - No TLS secret name specified by default!
+#### Default: `""` - No TLS secret name specified by default!
```
marcinc, line 988: Have we changes the way it's handled internally now? For things that can practically be defined at all maybe we should allow nil value and default to nil? I understand that the config struct now defines that as string hence this change?

marcinc (commented):
Left a bunch of minor comments and change suggestions. I think we also need to cover tutorials and examples with the latest changes.

docs/reference/config-example.md
```diff
@@ -10,59 +10,81 @@ version: "3.7"
...
-      kev.workload.readiness-probe-initial-delay: 1m0s    # Default: 1m. How long to wait before initial probe run. Shared by all readiness probe types.
-      kev.workload.readiness-probe-interval: 1m0s         # Default: 1m. Interval for the probe. Shared by all readiness probe types.
-      kev.workload.readiness-probe-retries: "3"           # Default: 3. Number of probe retires. Shared by all readiness probe types.
-      kev.workload.readiness-probe-timeout: 10s           # Default: 10s. Probe timeout. Shared by all readiness probe types.
+    x-k8s:                                               # compose K8s configuration extension
+      disabled: false                                    # Disables/enables running as a K8s service - defaults to false. 
+      service:                                           # K8s service configuration (only required if values are overridden)
+        type: None                                       # Default: none (no service). Possible options: none | headless | clusterip | nodeport | loadbalancer.
+        nodeport: ""                                     # Default: nil. Only taken into account when working with service.type: nodeport
```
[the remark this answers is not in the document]
ezodude: Nope, that should be a nil. Will fix.

docs/reference/config-example.md
```diff
@@ -10,59 +10,81 @@ version: "3.7"
...
+        expose:                                          # K8s configuration to expose a service (not required if defaults are to be used)
+          domain: ""                                     # Default: "" (no ingress). Possible options: "" | domain.com,otherdomain.com (comma separated domain names). When true / domain(s) - it'll set ingress object.
+          tlsSecret: ""                                  # Default: "" (no tls). Secret name where certs will be loaded from.
```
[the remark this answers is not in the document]
ezodude: That was a 1 to 1 mapping from the old one. Will change.

docs/reference/config-example.md
```diff
@@ -10,59 +10,81 @@ version: "3.7"
...
-      kev.workload.readiness-probe-timeout: 10s           # Default: 10s. Probe timeout. Shared by all readiness probe types.
+    x-k8s:                                               # compose K8s configuration extension
+      disabled: false                                    # Disables/enables running as a K8s service - defaults to false. 
+      service:                                           # K8s service configuration (only required if values are overridden)
+        type: None                                       # Default: none (no service). Possible options: none | headless | clusterip | nodeport | loadbalancer.
+        nodeport: ""                                     # Default: nil. Only taken into account when working with service.type: nodeport
+        expose:                                          # K8s configuration to expose a service (not required if defaults are to be used)
+          domain: ""                                     # Default: "" (no ingress). Possible options: "" | domain.com,otherdomain.com (comma separated domain names). When true / domain(s) - it'll set ingress object.
```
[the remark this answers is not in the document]
ezodude: Does true, still work (my bad for leaving it in)? I thought it only took comma separated domain names only.

But the rewording makes sense.

docs/reference/config-example.md
```diff
@@ -10,59 +10,81 @@ version: "3.7"
...
+        autoscale:                                       # Configures an application for auto-scaling.
+          maxReplicas: 0                                 # Default: 0. Number of replicas to autoscale to.
+          cpuThreshold: 70                               # Default: 70. The CPU utilisation threshold.
+          memThreshold: 70                               # Default: 70. The Memory utilization threshold.
+        imagePull:                                       # The docker image pull policy.
+          policy: IfNotPresent                           # Default: IfNotPresent. Possible options: IfNotPresent / Always.
+          secret: ""                                     # Default: "" (no secret). Docker image pull secret to pull images from the container registry.
+        livenessProbe:                                   # Workload's liveness probe
+          ### EXEC
+          type: exec                                     # Default: exec. Possible options: none | exec | http | tcp.
```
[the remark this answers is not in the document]
ezodude: Again, just copied this one over. But rewording makes more sense.

docs/reference/config-params.md
ezodude: That one was missed - thanks!

docs/reference/config-params.md
ezodude: `retries` has been covered with `failureThreshold`.

docs/reference/config-params.md
ezodude: Yup, internally, it's a string.

Will revert back to `nil` ... this should work with strings too.

docs/reference/config-example.md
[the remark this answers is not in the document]
marcinc: Just double checked and it looks like that's correct - it's either empty string (which means default ingress will be created), or a comma separated list of domains. What's the default value for `expose.domains` if this wasn't defined in source compose and env overrides?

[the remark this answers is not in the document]
ezodude: Updated, but removed wording relating to `true`.

docs/reference/config-params.md
ezodude: Agreed, not really required as it will become obvious.

docs/reference/config-params.md
ezodude: Updated to: `echo "<generic prompt text>" ` as we don't care what the actual text is.

docs/reference/config-params.md
ezodude: `interval` is now `period`.

docs/reference/config-params.md
ezodude: `interval` is now `period`.

docs/reference/config-params.md
ezodude: They're on par.

docs/reference/config-params.md
ezodude: As discussed, `true` will be treated as a new feature in upcoming ticket.

ezodude:
Also, added the new `successThreshold` (liveness/readiness probe) param.

docs/reference/config-example.md
```diff
@@ -10,59 +10,83 @@ version: "3.7"
...
-      kev.workload.readiness-probe-interval: 1m0s         # Default: 1m. Interval for the probe. Shared by all readiness probe types.
-      kev.workload.readiness-probe-retries: "3"           # Default: 3. Number of probe retires. Shared by all readiness probe types.
-      kev.workload.readiness-probe-timeout: 10s           # Default: 10s. Probe timeout. Shared by all readiness probe types.
+    x-k8s:                                               # compose K8s configuration extension
+      disabled: false                                    # When disabled the compose service won't be included in generated K8s manifests. Defaults to 'false'. 
+      service:                                           # K8s service configuration (only required if values are overridden)
+        type: None                                       # Default: none (no service). Possible options: none | headless | clusterip | nodeport | loadbalancer.
+        nodeport:                                        # Default: nil. Only taken into account when working with service.type: nodeport
```
marcinc, line 17: It's good. Perhaps we should give an example in the comment? e.g. `Set with numeric value e.g. 5555` or something like that?

docs/reference/config-example.md
```diff
@@ -10,59 +10,83 @@ version: "3.7"
...
+        expose:                                          # K8s configuration to expose a service externally (by default services are never exposed)
+          domain:                                        # Default: "" (no ingress). Possible options: "" | domain.com,otherdomain.com (comma separated domain names). When with domain name(s) - it'll generate an ingress object and expose service externally.
```
marcinc, line 19: OK, so how do we expose service via ingress with default backend? Setting domain value to "default"? If so we should perhaps include that in the comment?

docs/reference/config-params.md
marcinc: I guess this hasn't been addressed? I think it'd be useful to provide some context here, especially because we present example containing more than one setting that this section is dedicated to.

docs/reference/config-params.md
marcinc: ok, I guess the docs change related to that will be also in the upcoming PR?

marcinc (commented):
A few more minor comments. The examples / tutorials will need to reflect reference changes too.

docs/reference/config-params.md
ezodude: Will fix now as its been finalised.

ezodude:
Yup, I somehow missed the examples. Will fix now too.

ezodude:
@marcinc I think I need more clarification for: https://github.com/appvia/kev/pull/547#discussion_r644832114

Not sure I understand what extra context is needed?

marcinc (approved):
LG 🎉

PR #548, opened 2021-06-01
Set correct livenessProbe success threshold

### LINKED ISSUE #544: Set correct livenessProbe success threshold
Currently we don't distinguish the configuration of different probe types. 

Liveness probe `successTheshold` throws exception on value higher than `1`.

For liveness and startup probes the value of successThreshold must be set to 1.

See [official documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes)

"successThreshold: Minimum consecutive successes for the probe to be considered successful after having failed. Defaults to 1. Must be 1 for liveness and startup Probes. Minimum value is 1."

ezodude:
Resolves #544 

- Introduces success threshold config for a `v1Probe`.
- Ensures the `LivenessProbe` success threshold is always set to `1`.

### DIFF
--- a/pkg/kev/config/defaults.go
+++ b/pkg/kev/config/defaults.go
@@ -79,9 +79,6 @@ const (
 	// Default: 0.1, which is equivalent to 10% of CPU
 	DefaultResourceRequestCPU = "0.1"
 
-	// DefaultLivenessProbeType default probe type.
-	DefaultLivenessProbeType = "exec"
-
 	// DefaultProbeTimeout default 10s
 	DefaultProbeTimeout = "10s"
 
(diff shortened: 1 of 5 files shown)


pkg/kev/config/defaults.go
```diff
@@ -91,9 +91,12 @@ const (
 	// DefaultProbeInitialDelay default 1m (1 minute)
 	DefaultProbeInitialDelay = "1m"
 
-	// DefaultProbeRetries default 3. Number of retries for liveness probe command
+	// DefaultProbeRetries default 3. Number of retries for probe command
 	DefaultProbeRetries = 3
 
+	// DefaultSuccessThreshold default 1. Minimum consecutive successes for the probe to be considered successful
+	DefaultSuccessThreshold = 1
```
marcinc, line 98: small rename suggestion: `DefaultSuccessThreshold` -> `DefaultProbeSuccessThreshold` to keep it in line with the other probe defaults.

marcinc: Actually on a second thought perhaps we shouldn't place a default here as there is no equivalent for this in Compose spec. We can simply assume it's 1 by default. 

Instead we should consider adding a dedicated config key for probe success threshold in an extension driven config and have it override the default in applicable probe types.

marcinc (approved):
LGTM

PR #552, opened 2021-06-03
Handle ingress with kubernetes cluster defaults

### LINKED ISSUE #550: Handle ingress with kubernetes cluster defaults
Simplify Ingress creation logic: 

1) If the `default` keyword is provided, instead of a list of domains: we create an ingress without host information (i.e. default backend).

2) If domain(s) are specified - we generate ingress object with host itemised.

ezodude:
Resolves #550

- Exposed service with keyword `default` will use the default ingress backend.
- Exposed service with keyword `default` will not create a TLS ingress spec as per [docs](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls).

### DIFF
--- a/pkg/kev/config/service.go
+++ b/pkg/kev/config/service.go
@@ -88,7 +88,7 @@ func inferServiceTypeFromComposeValue(v string) (ServiceType, error) {
 	case "nodeport":
 		return NodePortService, nil
 	case "loadbalancer":
-		return NodePortService, nil
+		return LoadBalancerService, nil
 	case "headless":
 		return HeadlessService, nil
 	case "none":
--- a/pkg/kev/converter/kubernetes/transform.go
+++ b/pkg/kev/converter/kubernetes/transform.go
@@ -51,6 +51,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const DefaultIngressBackendMarker = "default"
+
 // Kubernetes transformer
 type Kubernetes struct {
 	Opt      ConvertOptions     // user provided options from the command line
@@ -153,7 +155,7 @@ func (k *Kubernetes) Transform() ([]runtime.Object, error) {
 			}
 			objects = append(objects, svc)
 
-			// For exposed service also create an ingress (Note only the first port is used for ingress!)
+			// For exposed service also create an ingress (Note: only the first port is used for ingress!)
 			expose, err := projectService.exposeService()
 			if err != nil {
 				msg := "Could not expose the service. Ingress hasn't been created"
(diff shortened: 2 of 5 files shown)


pkg/kev/config/service.go
```diff
@@ -88,7 +88,7 @@ func inferServiceTypeFromComposeValue(v string) (ServiceType, error) {
 	case "nodeport":
 		return NodePortService, nil
 	case "loadbalancer":
-		return NodePortService, nil
+		return LoadBalancerService, nil
```
marcinc, line 91: Good catch! 👍

pkg/kev/converter/kubernetes/utils.go
```diff
@@ -950,3 +952,31 @@ func volumeByNameAndFormat(name string, formatter func(string) string, volumes c
 	}
 	return composego.VolumeConfig{}
 }
+
+// hasDefaultIngressBackendKeyword determines whether the host value list contains the keyword used to create
+// a default backend ingress.
+func hasDefaultIngressBackendKeyword(v []string) bool {
+	return strings.Contains(strings.Join(v, ""), DefaultIngressBackendKeyword)
```
marcinc, line 959: should we also strip whitespaces here or is it done by the config object already?

pkg/kev/converter/kubernetes/transform_test.go
```diff
@@ -884,6 +896,18 @@ var _ = Describe("Transform", func() {
...
 			})
 		})
+
+		When("TLS secret name was specified via extension for service exposed with default ingress backend", func() {
+			BeforeEach(func() {
+				projectService.SvcK8sConfig.Service.Expose.Domain = DefaultIngressBackendKeyword
+				projectService.SvcK8sConfig.Service.Expose.TlsSecret = "my-tls-secret"
+			})
+
+			It("does not create a TLS object it in the ingress spec", func() {
```
marcinc, line 906: `object it in` -> `object in`

pkg/kev/converter/kubernetes/transform.go
```diff
@@ -645,39 +647,26 @@ func (k *Kubernetes) initIngress(projectService ProjectService, port int32) *net
...
-						},
-					},
-				},
-			},
-		}
-		if host != "true" {
-			ingress.Spec.Rules[i].Host = host
+	if hasDefaultIngressBackendKeyword(hosts) {
+		ingress.Spec.Rules = []networkingv1beta1.IngressRule{
+			createIngressRule("", "", projectService.Name, port),
```
marcinc, line 659: I wonder whether this is going to result with a valid ingress path? Perhaps "/" as path would be more appropriate? Worth double checking with the docs for this one.

marcinc (commented):
a few minor comments

pkg/kev/converter/kubernetes/transform.go
ezodude: Thanks for the catch! And, **you shouldn't need a path**. In fact the implementation was wrong.

The default backend should not produce any rules. It should create an ingress object similar to this:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-ingress
spec:
  defaultBackend:
    service:
      name: test
      port:
        number: 80
```

I've corrected the implementation to do the above using the `*networkingv1beta1.IngressBackend` struct.

pkg/kev/converter/kubernetes/utils.go
ezodude: Good catch!

Will trim space in the `p.SvcK8sConfig.Service.Expose.Domain` value in `exposeService()` before it's returned.

marcinc: I think it'd make sense to do it in this PR

marcinc (commented):
Changes look good. Left one comment re stripping whitespaces.

pkg/kev/converter/kubernetes/utils.go
ezodude: @marcinc The whitespace is stripped upstream - there's no extra whitespace at this point.

But maybe I missing the point?

ezodude:
@marcinc see remarks here: https://github.com/appvia/kev/pull/552#discussion_r645453452

Not sure what's still outstanding?

marcinc (approved):
👍

PR #561, opened 2021-06-04
Ensures file paths are build consistently across the codebase

### LINKED ISSUE #560: Ensure all filesystem paths are build correctly
We currently use a mix of `path.Join` and `filepath.Join`. Ensure paths are always correctly resolved by using `filepath.Join` across the board.

marcinc:
Resolved #560 

This PR makes all file paths to be built consistently across the codebase.


ezodude (approved):
👍 

Had to understand the difference between using one over the other. Makes sense.

PR #565, opened 2021-06-11
Clean up and integrate logs with the CLI UI.

### LINKED ISSUE #523: Clean up and integrate logs with the CLI UI.
In the past, logs were used to present a simple user interface.

The CLI has moved on. Now, these logs break the komando based UI when they are presented.

Logging to present info to a user interface should now be removed.

Debug logging should be displayed by the new UI.

ezodude:
Resolves #523 

- Reworks verbose mode to display logs integrated with UI sections and steps.
- Ensures there are no Info logs in the app as these have been overtaken by the UI.


pkg/kev/log/log.go
```diff
@@ -36,7 +36,7 @@ const (
 	errorDetailPrefix = "⨯ "
 
 	// detailPrefix debug log level prefix for info, debug, error, using unicode to ensure a visual on screen.
-	detailPrefix = " → "
+	// detailPrefix = " → "
```
marcinc, line 39: leftover line

pkg/kev/log/log.go
```diff
@@ -235,8 +235,8 @@ func decorate(level string, f ...Fields) logrus.Fields {
 		switch level {
 		case "debug-title":
 			fields["prefix"] = successTitlePrefix
-		case "debug":
-			fields["prefix"] = detailPrefix
+		// case "debug":
```
marcinc, line 238: logging levels probably need a review?

pkg/kev/project.go
```diff
@@ -246,3 +295,9 @@ func WithExcludeServicesByEnv(c map[string][]string) Options {
 		cfg.ExcludeServicesByEnv = c
 	}
 }
+
+func WithLogVerbose(c bool) Options {
```
marcinc, line 299: could you add a simple comment to this func?

pkg/kev/types.go
```diff
@@ -39,6 +40,7 @@ type runConfig struct {
 	// ExcludeServicesByEnv is used to exclude an environment's set of services from processing.
 	// Primary use is during render.
 	ExcludeServicesByEnv map[string][]string
+	LogVerbose           bool
```
marcinc, line 43: A simple comment would be useful

cmd/kev/cmd/dev.go
```diff
@@ -127,7 +127,7 @@ func init() {
 	rootCmd.AddCommand(devCmd)
 }
 
-func runDevCmd(cmd *cobra.Command, args []string) error {
+func runDevCmd(cmd *cobra.Command, _ []string) error {
 	skaffold, err := cmd.Flags().GetBool("skaffold")
```
marcinc, line 131: It's not the scope of this PR necessarily but I think we could fix it as we touch commands anyway... either handle err for each parameter getter, or just skip that altogether, and err check below in line 139? This applies to this and other commands where we don't handle errors in place.

marcinc (commented):
a few comments.

cmd/kev/cmd/dev.go
```diff
@@ -132,9 +132,9 @@ func runDevCmd(cmd *cobra.Command, _ []string) error {
 	namespace, err := cmd.Flags().GetString("namespace")
 	kubecontext, err := cmd.Flags().GetString("kubecontext")
 	kevenv, err := cmd.Flags().GetString("kev-env")
-	tail, _ := cmd.Flags().GetBool("tail")
-	manualTrigger, _ := cmd.Flags().GetBool("manual-trigger")
-	verbose, _ := cmd.Root().Flags().GetBool("verbose")
+	tail, err := cmd.Flags().GetBool("tail")
+	manualTrigger, err := cmd.Flags().GetBool("manual-trigger")
+	verbose, err := cmd.Root().Flags().GetBool("verbose")
```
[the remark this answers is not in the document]
ezodude: Yup agreed. This was supposed to be a poor man's error check.

I'll remove error checks for everything as,
- The cmd runners will error for invalid values.
- The Cobra CLI lib will error for mistyped/unknown flags.

ezodude:
Regarding `log.go`, I've reverted back to the point in history before we started utilising logs for UI display.

Removing all the detail related funcs.

And, turned the prefixes into exported vars as you suggested.

marcinc (approved):
👍  LG

Issue #564, opened 2021-06-10
Introduce new resource (ephemeral storage) requests / limits


marcinc:
We currently support CPU & Memory resource requests and limits for the workload. 

Add ephemeral storage requests / limits support, as described by [official docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#local-ephemeral-storage)

PR #566, opened 2021-06-11
Add storage resource params

marcinc:
Resolves: #564 

This PR adds the additional configuration parameters around workload resource requests/limits. Specifically for ephemeral storage that workload can reserve and its upper limit:

```yaml
x-k8s:
  workload:
    resource:
      storage: 1G
      maxStorage: 2G
```

You can express storage as a plain integer or as a fixed-point number using one of these suffixes: E, P, T, G, M, K. You can also use the power-of-two equivalents: Ei, Pi, Ti, Gi, Mi, Ki.

By default `storage`/`maxStorage` aren't specified and won't be included. Only when explicitly defined those parameters will be included in resource requests and limits.

### DIFF
--- a/pkg/kev/converter/kubernetes/project_service.go
+++ b/pkg/kev/converter/kubernetes/project_service.go
@@ -242,14 +242,18 @@ func (p *ProjectService) placement() map[string]string {
 }
 
 // resourceRequests returns workload resource requests (memory & cpu)
-// It parses CPU & Memory as k8s resource.Quantity regardless
+// It parses CPU, Memory & Ephemeral Storage as k8s resource.Quantity regardless
 // of how values are supplied (via deploy block or an extension).
+// Note: Only CPU & Memory requests can set via docker compose deploy block!
+//       Storage can only be set via extension parameter.
 // It supports resource notations:
 // - CPU: 0.1, 100m (which is the same as 0.1), 1
 // - Memory: 1, 1M, 1m, 1G, 1Gi
-func (p *ProjectService) resourceRequests() (*int64, *int64) {
+// - Storage: 128974848, 10M, 100Mi, 1G, 2Gi
+func (p *ProjectService) resourceRequests() (*int64, *int64, *int64) {
 	var memRequest int64
 	var cpuRequest int64
+	var storageRequest int64
 
 	// @step extract requests from deploy block if present
 	if p.Deploy != nil && p.Deploy.Resources.Reservations != nil {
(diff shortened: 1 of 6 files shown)


pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -242,14 +242,18 @@ func (p *ProjectService) placement() map[string]string {
 }
 
 // resourceRequests returns workload resource requests (memory & cpu)
-// It parses CPU & Memory as k8s resource.Quantity regardless
+// It parses CPU, Memory & Ephemeral Storage as k8s resource.Quantity regardless
 // of how values are supplied (via deploy block or an extension).
+// Note: Only CPU & Memory requests can set via docker compose deploy block!
```
ezodude, line 247: `can set via docker compose` -> `can be set via a docker compose`

pkg/kev/converter/kubernetes/project_service.go
```diff
@@ -268,18 +272,27 @@ func (p *ProjectService) resourceRequests() (*int64, *int64) {
...
+	}
+
+	return &memRequest, &cpuRequest, &storageRequest
 }
 
 // resourceLimits returns workload resource limits (memory & cpu)
-// It parses CPU & Memory as k8s resource.Quantity regardless
+// It parses CPU, Memory & Ephemeral Storage as k8s resource.Quantity regardless
 // of how values are supplied (via deploy block or an extension).
+// Note: Only CPU & Memory requests can set via docker compose deploy block!
```
ezodude, line 286: `can set via docker compose` -> `can be set via a docker compose`

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -657,9 +658,11 @@ var _ = Describe("ProjectService", func() {
 
 			When("not specified via extension", func() {
 				It("returns resource request as defined in deploy block", func() {
-					mem, cpu := projectService.resourceRequests()
+					mem, cpu, storage := projectService.resourceRequests()
 					Expect(*mem).To(BeEquivalentTo(1000))
 					Expect(*cpu).To(BeEquivalentTo(100))
+					// storage can't be specified in docker compose deploy block
```
ezodude, line 664: Not sure we need the comment as it's noted on the actual function itself.

pkg/kev/converter/kubernetes/project_service_test.go
```diff
@@ -712,9 +727,11 @@ var _ = Describe("ProjectService", func() {
 
 			When("not specified via extension", func() {
 				It("returns resource limit as defined in deploy block", func() {
-					mem, cpu := projectService.resourceLimits()
+					mem, cpu, storage := projectService.resourceLimits()
 					Expect(*mem).To(BeEquivalentTo(1000))
 					Expect(*cpu).To(BeEquivalentTo(100))
+					// storage can't be set via docker compose deploy block
```
ezodude, line 733: Not sure we need the comment as it's noted on the actual function itself.

ezodude (commented):
Ephemeral storage - we really needed this. Glad we have it now.

Left some comments.

ezodude (approved):
💯

PR #574, opened 2021-06-17
Bug fix: reconcile a "create" change with a new extension map.

### LINKED ISSUE #573: Changes in docker compose file not applied to all environment overrides
### Short issue description
Running through the [Getting Started](https://github.com/appvia/kev/blob/master/docs/tutorials/getting-started-with-kev.md) tutorial and updating the `docker-compose.yaml` to include the MySQL DB service and associated volume, I expected to see corresponding changes to the all compose override files (`local` and `stage` as per the tutorial) and generated k8s files after running `kev render`. The render step only updated the `dev` override + k8s files, and not `local` and `stage` - these overrides were missing service details for the DB, and StatefulSet/PersistentVolumeClaim files for k8s.

Kev version:
OS (Linux/Mac/Windows..): Mac OS 11.4
Docker Compose version:1.29.2
Kubernetes version: 1.21
Kubernetes distribution (e.g. GKE, EKS, ...): Docker Desktop

### How can we reproduce the issue?
1. Run through the tutorial up to adding the DB service, and make the required MySQL/volume changes to `docker-compose.yaml`
2. Run `kev render`
3. Check reconciled compose override files and generated k8s files - only the dev overrides are initially updated
4. Running `kev render` again will reconcile `local`, but leave `stage`
5. Running `kev render` a final time will reconcile `stage`

ezodude:
Resolves #573 

## Bug summary

During reconciliation, the base docker compose extensions map used to detect changes was mistakenly mutated.

This was not an issue if a single environment override was being reconciled.

But, it definitely was a problem if multiple environment overrides were reconciled at once as the first reconcile breaks the following reconcile.

## Bug fix

Ensure the underlying extensions map is not mutated and that a fresh one is created for reconciles "create" change.

Added tests to cover the regression.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -541,8 +541,6 @@ github.com/go-playground/locales v0.13.0 h1:HyWk6mgj5qFqCT5fjGBuRArbVDfE4hi8+e8c
 github.com/go-playground/locales v0.13.0/go.mod h1:taPMhCMXrRLJO55olJkUXHZBHCxTMfnGwq/HNwmWNS8=
 github.com/go-playground/universal-translator v0.17.0 h1:icxd5fm+REJzpZx7ZfpaD876Lmtgy7VtROAbHHXk8no=
 github.com/go-playground/universal-translator v0.17.0/go.mod h1:UkSxE5sNxxRwHyU+Scu5vgOQjsIJAF8j9muTVoKLVtA=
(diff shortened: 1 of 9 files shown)


chriswalker (approved):
Obviously lacking a fair bit of context on the code at the moment, but LGTM based on conversation in standup, and test changes all look good 👍

marcinc (approved):
LGTM

Issue #575, opened 2021-06-18
`kev dev` command doesn't render environments as expected


marcinc:
There is a bug in `kev dev` command whereby it wrongly handles environments for which the K8s manifest should be automatically rendered.

The above makes the development loop to only take into account the `dev` environment. 

Expected behaviour is described by the 2 scenarios below:

1) `kev dev` (without Skaffold loop enabled) should take all tracked environments into account by default and render all envs K8s manifest upon change to any of the docker compose override files.

2) `kev dev` (with Skaffold loop activated via `--skaffold`) should only ever render the K8s manifests for a specified environment (selected by `--kev-env` flag), or `dev` environment if none was specified. The chosen environment name also selects the K8s manifest that should be automatically deployed to a target Kubernetes cluster.

PR #576, opened 2021-06-18
Dev command environments handling fix

marcinc:
Resolves: #575 

This PR fixes the way `dev` command handles project environments in such a way that:

1) in standard development loop (without Skaffold) it takes all tracked environments into account by default. It renders all envs K8s manifest upon change to any of the environment specific docker compose override files.

2) in development loop (with Skaffold activated) it only renders the K8s manifests for a specified environment (selected by `--kev-env` flag), or `dev` environment if none was specified. Specified environment's manifests are auto-deployed to a target K8s cluster.

It also fixes `manifest#UpdateBuildArtifacts` so that slices comparison works as expected. NOTE: `reflect.DeepEqual` treats nil and empty slice as non equal!

### DIFF
--- a/pkg/kev/dev.go
+++ b/pkg/kev/dev.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"os"
 	"os/signal"
+	"regexp"
 	"syscall"
 
 	"github.com/appvia/kev/pkg/kev/log"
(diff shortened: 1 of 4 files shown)


pkg/kev/dev.go
```diff
@@ -143,11 +153,14 @@ func (r *DevRunner) Run() error {
 				kmd.WithStyle(kmd.LogStyle),
 			)
 
+			match := envRe.FindStringSubmatch(ch)
```
ezodude, line 156: I like how we're capturing the environment.

Should we also handle what to do in the case of non matches?

ezodude (commented):
Makes sense. One quick comment.

pkg/kev/dev.go
marcinc: We technically shouldn’t need to, only the files that are related to tracked environments are watched, so the chance that there won’t be a match is basically non-existent.

ezodude (approved):
👍

chriswalker (approved):
LGTM and clarifies my initial question that prompted this!

Issue #583, opened 2021-06-21
Generate ServiceAccount manifest when service account name specified


marcinc:
User can specify a service account for their workload, however, program doesn't currently generate a manifest for it. 

Add SA manifest generator so that there is one less step to get going when manifests are applied to the Kubernetes cluster.

PR #586, opened 2021-06-22
Generate ServiceAccount manifest for eligible objects if service account name specified

marcinc:
Resolves: #583 

When workload's k8s configuration specifies a service account name other than `default` it'll result in ServiceAccount manifest being generated when `render` command is executed. 

It's a NOOP for `default` service account name.


ezodude (commented):
Looks ace with one v small nitpick.

ezodude (approved):
👍

Issue #584, opened 2021-06-21
Review and change the naming convention of the artefacts produced by the program


marcinc:
Consider making changes to the files naming convention to make it more meaningful, and still reduce naming conflicts as much as possible.

chriswalker:
Plan is to change from (for example) `docker-compose.kev.dev.yaml` and so on to `docker-compose.env.dev.yaml`. This will require changes to:

* All test fixture filenames
* Documentation (tutorials and README

`kev.yaml` will also change names, but this is still under discussion.

marcinc:
Resolved by #588

PR #588, opened 2021-06-23
Rename application-generated filenames

chriswalker:
Addresses #584. 

Modifies generation of environment override filenames by providing a dedicated infix constant (rather than deriving the infix portion of the generated `docker-compose` filenames from the app manifest name, which is now going to deviate anyway). Specifically, now generates `docker-compose.env.[environment].yaml` instead of `docker-compose.kev.[environment].yaml`.

Also involves changes to:

* All tests and fixtures to use new naming conventions,
* The Wordpress example files, 
* The tutorial files to reflect new filename changes,
* The e2e help Bash script.

Raised as a draft initially as further work is incoming around renaming of `kev.yaml` to... something else, which will also be handled as part of this PR.


pkg/kev/manifest.go
```diff
@@ -34,6 +34,10 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// Infix component of generated docker-compose filenames
+// (e.g. results in docker-compose.env.dev.yaml)
+const dockerComposeInfix = "env"
```
marcinc, line 39: As this infix is only ever used in the environment override files how about renaming this const to `envOverrideFileInfix`?

marcinc (commented):
One tiny comment, otherwise LGTM 👍

ezodude (commented):
Looks good.

pkg/kev/manifest.go
chriswalker: Good call, done!

cmd/kev/cmd/init.go
```diff
@@ -25,7 +25,7 @@ var initLongDesc = `Tracks compose sources & creates deployment environments.
 
 Examples:
 
-  ### Initialise kev.yaml with root docker-compose.yml and override file tracking. Adds a sandbox dev deployment environment.
+  ### Initialise appmeta.yaml with root docker-compose.yml and override file tracking. Adds a sandbox dev deployment environment.
```
marcinc, line 28: I think we could just say that we are initialising the project with docker-compose.yam…

docs/tutorials/getting-started-with-kev.md
```diff
@@ -139,23 +139,23 @@ Kev has now been initialised and configured. It has,
 - Created `dev` (a sandbox used by Kev for continuous development), `local` (useful for testing on our own machine) and `staging` (useful for testing on a remote machine) _Compose environment overrides_.
 
 It has also generated four files:
-- `kev.yaml`, a manifest that describes our _source application definition_ and _Compose environment overrides_.
-- Three `docker-compose.kev.*.yaml` files to represent our _Compose environment overrides_.
+- `appmeta.yaml`, a manifest that describes our _source application definition_ and _Compose environment overrides_.
+- Three `docker-compose.env.*.yaml` files to represent our _Compose environment overrides_.
 
-#### Manifest: kev.yaml
+#### Manifest: appmeta.yaml
```
marcinc, line 145: Could we rephrase the tile of this section?

marcinc (commented):
Couple of tiny comments. Otherwise it’s good to go! 👍

cmd/kev/cmd/init.go
chriswalker: Done

docs/tutorials/getting-started-with-kev.md
chriswalker: Hm. If we rephrase this, we should also rephrase the following section header, which states the override files. Suggest maybe `Application metadata file` and `Compose environment override files` (and drop the filenames from the section headings) - then I'll reference the filenames in the section text

marcinc (approved):
LGTM! 👍

PR #600, opened 2021-07-09
Required labels visibility

### LINKED ISSUE #393: Required labels visibility
liveness-probe and other configuration labels don't really need to be visible in an environment override. 

Defaults extracted from compose or assumed by Kev will be in place as they are now, however, user should still have a way to override behaviour if/when necessary. 

This will clean up the environment overrides slightly and should be a better experience overall.

Once UI elements are sorted we could also display an environment summary showing assumed defaults / values and a pointer to documentation describing how to override those.

ezodude:
Resolves #393 

- Minified environment overrides now only contain services replicas.
- Updated tests to the new reality.

### DIFF
--- a/pkg/kev/config/minify.go
+++ b/pkg/kev/config/minify.go
@@ -16,11 +16,26 @@
 
 package config
 
+import (
+	"reflect"
+)
+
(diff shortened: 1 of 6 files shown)


marcinc (approved):
LGTM

Issue #607, opened 2021-07-29
Fix symbolic environment variable handling


marcinc:
**Issue description:**

Currently environment variable symbolic value handling is preventing me from using special case keywords such as `secret`, `config`, `pod`, `container` as literal values for my environment variables.

**Expected behaviour:**

```
# the following should be treated as literal value

MY_VAR_A: secret.string
MY_VAR_B: config
MY_VAR_C: pod.foo
MY_VAR_D: container.bar
```

PR #608, opened 2021-07-29
Fixes environment variable symbolic value handling

marcinc:
Addresses #607 

This Bugfix PR changes the way environment variable symbolic values are handled, specifically:
* variables values that contain special case keyword such as `secret`, `config`, `pod`, `container` is treated as literal string unless it matches specific format that would indicate Kubernetes object should be referenced instead.  
* Adds validation to the expected format for symbolic values
* Returns explicit errors on malformed / unsupported symbolic values


chriswalker (approved):
👍 LGTM

chriswalker (dismissed):
Nice!

chriswalker (approved):

PR #610, opened 2021-08-04
Allow updates to images in override files

chriswalker:
Allows users to change the image names in override files, and persist these in the generated k8s manifests. This allows users to specify different image files for different environments (which otherwise cannot be done when deriving images from a single `docker-compose.yaml` alone.)

This change adds a new `Image` field to the `ServiceConfig` struct, and adds new Service update handling to the overrides/changesets. Test fixture files have been updated to reflect the fact they now always contain images for defined services.

### DIFF
--- a/pkg/kev/override.go
+++ b/pkg/kev/override.go
@@ -139,27 +139,28 @@ func (o *composeOverride) detectAndPatchServicesUpdate(dst *composeOverride) err
 	step := sg.Add("Detecting service updates")
 
 	cset := changeset{}
-	srcSvcSet := o.Services.Set()
 	for index, srcSvc := range o.Services {
-		if srcSvcSet[srcSvc.Name] {
-			dstSvc, err := dst.getService(srcSvc.Name)
-			if err != nil {
-				// Service not present in the override, which means this isn't
-				// an update; it should be handled by create checks
-				continue
-			}
-			// A note here about change semantic: usually the docker-compose file
-			// overrides the generate override files (e.g. deleting a service, volume
-			// etc). However, in the case of service updates,
-			if srcSvc.Image != dstSvc.Image {
-				cset.services = append(cset.services, change{
-					Type:   UPDATE,
-					Index:  index,
-					Parent: "services",
-					Value:  dstSvc.Image,
-				})
-				log.Debugf("detected an updated service image: %s", dstSvc.Image)
-			}
+		dstSvc, err := dst.getService(srcSvc.Name)
+		if err != nil {
+			// Service not present in the override, which means this isn't
+			// an update; it should be handled by create checks
+			continue
+		}
+		// A note here about change semantics: usually docker-compose file
+		// elements (such as ervices or volumes) influence the content of
+		// environment-specific overrides (i.e. changes in the docker-compose
+		// file take precedence over those in the overrides), However, in
+		// the case of service updates, this works the other way around: image
+		// changes in an override should be maintained over the original in the
+		// docker-compose file.
+		if srcSvc.Image != dstSvc.Image {
+			cset.services = append(cset.services, change{
+				Type:   UPDATE,
+				Index:  index,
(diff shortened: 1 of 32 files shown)


chriswalker:
There is an outstanding UI wrinkle with the above changes: because images in the override files take precedence over those defined in the `docker-compose.yaml` file, if they differ, this always gets reported as an update in the UI output. So, for example:

* Run `kev init` to generate project files, including overrides
* Edit the `dev` override file to use a different image to the one specified in the `docker-compose.yaml`, and run `kev render`
* This reports the service as being updated (correctly), and generates a correct `Image` entry in the resulting deployment manifest
* Run `kev render` again with no changes to the override - the UI reports the service as being updated (because the override is different to the originating value in the `docker-compose.yaml`), even though technically no changes have occurred.

Not currently sure how to tackle this (or even if we can, given we can't check differences between overrides on successive runs).

pkg/kev/override.go
```diff
@@ -311,9 +387,16 @@ func (o *composeOverride) mergeServicesInto(p *ComposeProject) error {
...
 			return err
 		}
-
 		envVarsFromNilToBlankInService(base)
 
+		// Copy over anything that is defined outside an extension or
+		// environment - users may modify some elements of the generated
+		// override files such as image names (for example, to provide
+		// different images for staging and prod environments)
+		if override.Image != base.Image {
```
marcinc, line 396: I think this works for now. Would it make sense to introduce an enum of things we want to respect or is it too early?

marcinc (commented):
A few minor comments. It'd be great to provide a test coverage for both cases 1) when environment override defines an image, and 2) when environment override doesn't override it at all. I don't think we need to change existing test assets too much btw.

One thing I noticed when testing was that when an `image` is dropped entirely / not defined in an env override it will be recreated as `image: ""` post render.  

It'd be best to leave environment override as is and not inject empty string image name.

To recreate: 

1) my dev env override config pre render:
```yaml
services:
  myService:
    x-k8s:
      ...
```

2) run `kev render`

3) dev env override config post render:

```yaml
services:
  myService:
    image: ""
    x-k8s:
      ...
```

I also see the following in the `render` CLI output, in "Detecting project updates" section

```sh
» Detecting project updates...
dev: config.env.dev.yaml
 ...
 ✓  Applied service updates
   | updated image to:
 ...
```

chriswalker:
Empty string for the image definitely a bug (which those tests would catch, for sure) - will fix that

pkg/kev/override.go
```diff
@@ -107,6 +133,56 @@ func (o *composeOverride) detectAndPatchVersionUpdate(dst *composeOverride) {
...
 
+func (o *composeOverride) detectAndPatchServicesUpdate(dst *composeOverride) error {
+	sg := o.UI.StepGroup()
+	defer sg.Done()
+	step := sg.Add("Detecting service updates")
+
+	cset := changeset{}
+	srcSvcSet := o.Services.Set()
+	for index, srcSvc := range o.Services {
+		if srcSvcSet[srcSvc.Name] {
```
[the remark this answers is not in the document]
chriswalker: Yeah, mistake and not needed here (plus doesn't make any sense - other tests check the `dstSvcSet[srcSvc.Name]` which _does_ make sense.) Will remove.

pkg/kev/override.go
chriswalker: Agree, fine for now, but will adjust if/when we start adding other fields to the 'precedence' list

chriswalker:
Added unit tests (fixtures under `testdata/init-image`), also modified comment you referenced in Slack so it's clearer (and finished, ahem)

chriswalker:
Closing as per discussion with @marcinc; a lot of the work done here is not actually needed (I now realise), so will redo this in a new branch/PR, which will have far fewer changes in it.

PR #617, opened 2021-08-16
Bump k8s ingress api to v1

### LINKED ISSUE #616: Update transform code to use v1 for networking API instead of v1beta
Current `pkg/kev/converter/kubernetes/transform.go` code uses v1beta for the Ingress components; update to v1 for this

marcinc:
Resolves #616 

This PR updates generated ingress object to `v1` from previous `v1beta1`.

### DIFF
--- a/docs/reference/config-params.md
+++ b/docs/reference/config-params.md
@@ -1114,7 +1114,9 @@ Defines how to expose the service externally. By default, all component services
 #### Possible options:
 
 - "default" - ingress will be created with Kubernetes cluster defaults.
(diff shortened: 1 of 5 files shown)


chriswalker (dismissed):
💥 LGTM

chriswalker (approved):
LGTM. General comment (and nothing to do with your changes), but note we're inconsistent with our package alias naming - `v1apps`, `networkingv1`, and then the `meta` alias doesn't have a version in it at all - minor thing to tweak when we have time. But like the fact we (mostly) reference the API version in the package alias, which is super-helpful when it comes to k8s APIs

PR #618, opened 2021-08-19
Allow image name overrides in environment override files

chriswalker:
Allow users to override the images in the environment override files, and persist these in the generated k8s manifests. This allows users to specify different images for different environments.

This change adds a new `Image` field to the `ServiceConfig` struct, and modifies the post-reconciliation merging to copy over override image names, if present. Unit tests are modeled on those checking env var overrides, and ensure that the reconciliation processing does not remove any specified image names in the overrides.

This PR replaces #610.

### DIFF
--- a/pkg/kev/override.go
+++ b/pkg/kev/override.go
@@ -314,6 +314,12 @@ func (o *composeOverride) mergeServicesInto(p *ComposeProject) error {
 
 		envVarsFromNilToBlankInService(base)
 
+		// Copy over image name if one has bee defined in the override. In
+		// future this may expand to invlude other fields.
+		if override.Image != "" && override.Image != base.Image {
+			base.Image = override.Image
+		}
+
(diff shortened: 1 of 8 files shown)


pkg/kev/override.go
```diff
@@ -314,6 +314,12 @@ func (o *composeOverride) mergeServicesInto(p *ComposeProject) error {
 
 		envVarsFromNilToBlankInService(base)
 
+		// Copy over image name if one has bee defined in the override. In
+		// future this may expand to invlude other fields.
+		if override.Image != "" && override.Image != base.Image {
```
marcinc, line 319: minor nitpick. It'd be better to trim whitespaces and check for length instead in this condition I think.

marcinc (commented):
1 minor comment. Otherwise LGTM.

marcinc (approved):
LGTM

PR #701, opened 2022-11-08
Add ability to specify prefix for exposed service hostname

marcinc:
This PR adds the ability to specify hostname prefix for the exposed service. 

When specified it'll be prepended to the domain name.

Example: 
 
For prefix `hello.` and domain name `world.example.com` it'll result in ingress hostname set as `hello.world.example.com`.

Example config:

```yaml
version: 3.7
services:
  my-service:
    x-k8s:
      service:
        expose:
          domainPrefix: hello.
          domain: world.example.com
...
```

### DIFF
--- a/pkg/kev/converter/kubernetes/transform_test.go
+++ b/pkg/kev/converter/kubernetes/transform_test.go
@@ -979,6 +979,19 @@ var _ = Describe("Transform", func() {
 			})
 		})
 
+		When("project servive extension is exposing the k8s service using a domain name and prefix", func() {
+			BeforeEach(func() {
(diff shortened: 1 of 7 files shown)


pkg/kev/converter/kubernetes/transform_test.go
```diff
@@ -979,6 +979,19 @@ var _ = Describe("Transform", func() {
 			})
 		})
 
+		When("project servive extension is exposing the k8s service using a domain name and prefix", func() {
```
KashifSaadat, line 982: `servive` => `service`

KashifSaadat (dismissed):
Nice 👍

KashifSaadat (approved):

PR #708, opened 2022-11-15
Skaffold config will preserve Sync sections if defined

marcinc:
Any file syncing rules that may have been manually added to the `skaffold.yaml` manifest will now be preserved for each respective build artefact. Previously `render` and `dev` subcommands would strip `sync` block entirely.

Example: 
```yaml
apiVersion: skaffold/v2beta14
kind: Config
metadata:
  name: App
build:
  artifacts:
    - image: myregistry.example.com/org/repo
      context: app
      sync:    # <---- manually added file syncing block will now be preserved with `kev render` / `kev dev` commands.
        manual:
          - src: src/**/*
            dest: /app
  tagPolicy:
    gitCommit:
      variant: Tags
  local: {}
```


KashifSaadat (approved):
LGTM

PR #736, opened 2023-02-09
[FEATURE] Patch command

marcinc:
This PR adds new `patch` subcommand that is responsible for patching images for specified services in previously generated K8s manifests. This might come handy when iterating on the application externally e.g. from CI, in particular with image building and publishing managed by external processes. 

NOTE: 
- By default all changes are inline i.e. original manifests for specified services will be overridden in place.
  - It's possible to provide output directory for patched manifests via `--output-dir` (Caveat: only patched manifests will be stored in the output dir!)
- Source directory is traversed and all matching manifests will be affected. In order to scope patching to a particular set of manifests make sure that the path is provided specifically to that one directory.

```
Examples:

### Patch images for specified services
$ tako patch --dir /path/to/k8s/manifests --image myservice=myimage:v1.0.1 --image myservice2=myimage2:v2.0.2

### Patch images for specified services and store patched manifests in a different directory
$ tako patch --dir /path/to/k8s/manifests --image myservice=myimage:v1.0.1 --image myservice2=myimage2:v2.0.2 --output-dir /path/to/patched/manifests

Usage:
  tako patch [flags]

Flags:
  -d, --dir string          Path to the kubernetes manifests that require patching. (default ".")
  -i, --image strings       Image to be patched in deployment files for a given service name. Format: <service>=<image>.
  -o, --output-dir string   Path to directory where patched manifests should be stored.
                            ⌙ Output directory structure will reflect that of the source directory tree.
                            ⌙ Note that only manifests that were patched will be stored in the output directory!
                            ⌙ Manifests will be overridden in-place if output directory is not specified.
  -h, --help                help for patch
  ```


KashifSaadat (dismissed):
LGTM

KashifSaadat (approved):

